PR #8, opened 2019-07-23
Make F# version more idiomatic.

isaacabraham:
Made some changes to the code that is perhaps a little more idiomatic F#:

* Switch to use of Records instead of Classes for the domain in `MyContracts.fs`. This allows better type inference e.g. `MyCalculator.fs` and simpler value construction e.g. `Client_FS\Program.fs`, and is the standard way of constructing data values in F# rather than Classes. Note the use of the `[<CLIMutable>]` attribute which enforces construction of setters at the runtime level (even though the types are still immutable from an F# point of view).
* Removed unnecessary bindings on arguments that are not used
* Removed unnecessary type annotations where type inference is sufficient
* Removed unnecessary `new` keyword where not required
* Some minor changes to formatting


mgravell:
Awesome, thanks; I used all of my F# know-how and concluded "yes, that looks like F#" :)

isaacabraham:
@mgravell yep, I know :-) It's great that you put this in there - happy to help try to make it even more awesome :-)

isaacabraham:
Out of curiosity, is it possible to control how to deserialization works? As in, is there somewhere that you use reflection to rehydrate objects or is this done at the protobuf layer? If we can control this, then we could remove the need for the `[<CLIMutable>]` attribute as well.

mgravell:
@isaacabraham the marshaller is 100% swappable, but if you mean "with protobuf-net", then my recommendation would be: "you tell me what an idiomatic/perfect F# contract would look like, and I try and make it work with protobuf-net natively". The one concern that I'd raise, however, is version-tolerance. For example, protobuf-net works fine with types that look like tuples (a constructor that covers every member, inferring the tag numbers from the constructor positions), but that tends to be brittle if someone can add a member *in the middle* later.

mgravell:
Note also: protobuf-net can see past Roslyn-style get-only members; if F# uses a different signature, we can educate it!

isaacabraham:
At runtime, an F# record is just a normal class. However, the main issue for many libraries is that without `CLIMutable`:

1. F# generates immutable objects - no public setters for any properties (F# record fields basically map to getter-only properties). There are obviously private backing fields for those properties, although (annoying) the naming convention is slightly different to C#s.
2. F# generates a single non-default constructor that takes in all fields. This is how it can "force" objects to never be created in a half-initialised state.

The only way to "rehydrate" such an object is either (AFAIK):

* Looking for the backing field with reflection and manually setting the value there, although you'll have to figure out some way to create the class instance without calling the constructor (or supply default values for them) and then immediately override them.

OR

* Call the non-default constructor, supply all the values in the correct position (I can't remember in what order the constructor args are - I think it's alphabetical - @cartermp can you confirm?)

mgravell:
Constructing types without the constructor is possible - the lib already supports that. But if we can figure out the constructor usage, it would probably be easier.

cartermp:
You can see what's generated here: https://sharplab.io/#v2:DYLgZgzgPgLgngBwKYAIBKKC8KDeKCCIKAlgHYwDcKAykRDAE5kDmKAvkA==

Relevant constructor:

```
        public R(int a, string s)
        {
            A@ = a;
            S@ = s;
        }
```

This is actually not accessible in F#, but it is accessible in C#. We're actually looking at making it accessible in F# as well.

PR #10, opened 2019-07-26
netcoreapp3.0 => netstandard2.1

pengweiqhca:


mgravell:
Question: why? What does this give you that the existing ns2.0 TFM doesn't?

pengweiqhca:
ns2.0 need reference `Microsoft.Bcl.AsyncInterfaces`

mgravell:
Ah, so your objective here is to retain the reduced dependency graph that nca3.0 gives. I'll need to take a look, but in principle, sure that makes sense.

mgravell:
Since Grpc.Net.Client targets ns2.1, this seems fine and reasonable; merging - thanks

PR #17, opened 2019-09-06
protobuf-net.Grpc.Test add net461 and net472

pengweiqhca:
`protobuf-net.Grpc` `TargetFrameworks` need add net461, `protobuf-net.Grpc.Test` net461 test failed
```
System.IO.FileNotFoundException
Cann't load assembly“System.ServiceModel.Primitives, Version=4.5.0.3, Culture=neutral, PublicKeyToken=b03f5f7f11d50a3a
```


mgravell:
This is a fun and complicated decision. Net 461 has a number of particularly gnarly binding problems, and in many ways shouldn't *actually* be considered netstandard2.0 compatible (yes, I know all about the matrix table; I also know that it is a simplification). Net 462 is a little less brittle, but frankly it is Net 472 that is the first version that should be considered "working".

So: as a library author, this puts me in a delicate position in terms of which TFMs to support:

- NS2.0 only
- NS2.0, Net 462 and Net 472 (the Net 472 is needed to **undo** the Net 462 bits)
- NS2.0, Net 461 and Net 472

Obviously *today* I have it set at the first. I'm *kinda* OK with the second. The third (Net 461) is less desirable to me, as I know it is going to cause brittle issues. Frankly, Net 461 needs to die.

So, question: why Net 461 specifically? If this is simply "because Net 461 lies and pretends to be NetStandard 2.0 compatible, meaning: you can compile a Net 461 project with the library", then: that isn't necessarily compelling. What is  your motivation for wanting a Net 461 target?

pengweiqhca:
we don't care about the user using net461, net462 or net472, we should only care about the user won't get `System.IO.FileNotFoundException` after reference protobuf-net.Grpc in netframework.

mgravell:
Hmmm. I can see the point, but it is kinda frustrating that net461 comes pre-broken in many ways *anyway*. But: we can't stop the tools from claiming (incorrectly) that it'll work :)

Could you add a net472 target too? This is kinda required to undo the hackery that net461-net471 do.

pengweiqhca:
I added net472

mgravell:
Thanks. I'm out until Tuesday for medical reasons, but: it is my intention to merge this.

Can I confirm that you're happy to contribute this PR in line with the project license etc.

pengweiqhca:
Yes

PR #23, opened 2019-10-09
Add alternative code-first POCO GRPC Services

mythz:
Hi Mark,

Not sure how open you are to allow for an additional alternative code-first POCO alternative to specifying code-first Services, but the WCF-style interface isn't friendly to code-gen or being able to build genericized typed clients over specialized interfaces.

So I've included an (IMO) more natural alternative for defining GRPC Services by just having non-attributed POCO classes implementing the empty `IGrpcService` marker interface, e.g:

```csharp
public class ApplyServices : IGrpcService
{
    public Task<ApplyResponse> Add(Apply request) => 
        Task.FromResult(new ApplyResponse(request.X + request.Y));
    public Task<ApplyResponse> Mul(Apply request) => 
        Task.FromResult(new ApplyResponse(request.X * request.Y));
    public Task<ApplyResponse> Sub(Apply request) => 
        Task.FromResult(new ApplyResponse(request.X - request.Y));
    public Task<ApplyResponse> Div(Apply request) => 
        Task.FromResult(new ApplyResponse(request.X / request.Y));
}
```

Where the CodeFirst support will auto-wire all public methods in the class (except for Object methods) as GRPC Services. This was fairly non-disruptive to support (just a few LOC's in Binder classes).

I've included an integration test project to show how this can work without interfaces on the client using your Unary extension methods.

Let me know if you're open to this or would like me to make any further changes.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,4 +1,5 @@
 .vs
+.idea/
 artifacts
 bin
 obj
--- a/protobuf-net.Grpc.sln
+++ b/protobuf-net.Grpc.sln
@@ -92,6 +92,8 @@ Project("{2150E333-8FDC-42A3-9474-1A3956D46DE8}") = "perf", "perf", "{6A3AC822-5
 EndProject
 Project("{9A19103F-16F7-4668-BE54-9A1E7A4F7556}") = "PerfClient", "examples\perf\PerfClient\PerfClient.csproj", "{BCB98A08-8A05-41CA-B42F-9DBF3DB8D546}"
 EndProject
+Project("{FAE04EC0-301F-11D3-BF4B-00C04F79EFBC}") = "protobuf-net.Grpc.Test.Integration", "tests\protobuf-net.Grpc.Test.Integration\protobuf-net.Grpc.Test.Integration.csproj", "{7AF5B934-AEE9-4FD1-928D-AAE0F7098A32}"
+EndProject
 Global
 	GlobalSection(SolutionConfigurationPlatforms) = preSolution
 		Debug|Any CPU = Debug|Any CPU
@@ -237,6 +239,12 @@ Global
 		{BCB98A08-8A05-41CA-B42F-9DBF3DB8D546}.Release|Any CPU.Build.0 = Release|Any CPU
 		{BCB98A08-8A05-41CA-B42F-9DBF3DB8D546}.VS|Any CPU.ActiveCfg = Debug|Any CPU
 		{BCB98A08-8A05-41CA-B42F-9DBF3DB8D546}.VS|Any CPU.Build.0 = Debug|Any CPU
+		{7AF5B934-AEE9-4FD1-928D-AAE0F7098A32}.Debug|Any CPU.ActiveCfg = Debug|Any CPU
+		{7AF5B934-AEE9-4FD1-928D-AAE0F7098A32}.Debug|Any CPU.Build.0 = Debug|Any CPU
+		{7AF5B934-AEE9-4FD1-928D-AAE0F7098A32}.Release|Any CPU.ActiveCfg = Release|Any CPU
+		{7AF5B934-AEE9-4FD1-928D-AAE0F7098A32}.Release|Any CPU.Build.0 = Release|Any CPU
+		{7AF5B934-AEE9-4FD1-928D-AAE0F7098A32}.VS|Any CPU.ActiveCfg = Debug|Any CPU
+		{7AF5B934-AEE9-4FD1-928D-AAE0F7098A32}.VS|Any CPU.Build.0 = Debug|Any CPU
 	EndGlobalSection
 	GlobalSection(SolutionProperties) = preSolution
 		HideSolutionNode = FALSE
@@ -275,6 +283,7 @@ Global
 		{419CBCB3-1FDC-4974-9F92-0E8BFEA3A7D7} = {6A3AC822-52E8-4AE2-8F71-385DF0D69891}
 		{6A3AC822-52E8-4AE2-8F71-385DF0D69891} = {F7FAC6AD-62B0-4B79-98AA-DBD99F84E4E9}
 		{BCB98A08-8A05-41CA-B42F-9DBF3DB8D546} = {6A3AC822-52E8-4AE2-8F71-385DF0D69891}
+		{7AF5B934-AEE9-4FD1-928D-AAE0F7098A32} = {0A84599D-2CE9-416E-888F-24652EEAB0B3}
 	EndGlobalSection
 	GlobalSection(ExtensibilityGlobals) = postSolution
 		SolutionGuid = {BA14B07C-CA29-430D-A600-F37A050636D3}
--- a/src/protobuf-net.Grpc/Configuration/IGrpcService.cs
+++ b/src/protobuf-net.Grpc/Configuration/IGrpcService.cs
@@ -1,0 +1,7 @@
+namespace ProtoBuf.Grpc.Configuration
+{
+    /// <summary>
+    /// Marker Interface to specify that all public methods on this class should be wired as GRPC Services 
+    /// </summary>
+    public interface IGrpcService {}
+}
--- a/src/protobuf-net.Grpc/Configuration/IGrpcService.cs
+++ b/src/protobuf-net.Grpc/Configuration/IGrpcService.cs
@@ -1,7 +1,9 @@
 namespace ProtoBuf.Grpc.Configuration
 {
     /// <summary>
-    /// Marker Interface to specify that all public methods on this class should be wired as GRPC Services 
+    /// When a type implements this interface, the type name (without any prefix or other qualification)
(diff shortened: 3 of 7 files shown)


mgravell:
Happy to consider options, but talk me through this a sec; we have the class shown above - fine. Now; how does this relate to the client and/or server? Or does it not at all, with it only being used to do the bindings?

Basically, given the class above:

- if I want to create a server, what do I do (do I just register that type itself?)
- if I want to create a client (proxy), what do I do?

I tried to from it from the code, but it wasn't clear to me, which is probably the starting point for API additions.

But in principle: "fine"

mythz:
> how does this relate to the client and/or server?

So `ApplyServices` defines the server implementation, an instance of which would be resolved from the IOC and executed as normal. 

> if I want to create a server, what do I do (do I just register that type itself?)

Users can register their implementations individually, or you could also scan for all types in an assembly that implements `IGprcService` and autowire them automatically, e.g:

```csharp
services.AddCodeFirst(typeof(ApplyServices).Assembly);
```

Although scanning the assembly and automatically registering types individually is also easy enough to do in the consuming server library. I'm personally code-gen'ing a class that inherits from a base class that implements `IGprcService` with all the methods I want to make available via GRPC and delegating that to a controller that executes the requested Service. Effectively this is approach the easiest way to "glue" GRPC APIs to internal/existing implementations.

> if I want to create a client (proxy), what do I do?

By allowing Message Request/Response Types users will be able to implement their own typed proxy client that wraps their own client/server conventions. E.g. I've created a `GrpcServiceClient` [typed Service Client](https://docs.servicestack.net/csharp-server-events-client) that lets you send messages using HTTP/REST Verbs Get/Post/Put/Delete/Patch, e.g:

```csharp
var client = new GrpcServiceClient(url);
var response = await client.GetAsync(new Apply { ... });
```

Without relying on server conventions you could create client that lets you call these POCO Services like:

```csharp
var client = new GrpcClient(url);
var response = await client.SendAsync<ApplyResponse>("Add", new Apply { ... });
```

A benefit of this approach is the Client/Server doesn't have to share any interfaces, with just the messages the server can freely extend GRPC APIs that existing clients will be able to call by just calling it by name and the same generic clients for calling any GRPC Services on any POCO GRPC Server, i.e. you don't need a separate typed client proxy that's specialized to a single interface.

mgravell:
So if I restate:

- for server scenarios: the type shown *is* the server
- for client scenarios: mostly it is useful for extension purposes, not for direct consumer usage (i.e. it may need extra plumbing by something like ServiceStack)

Is that right? if so, I think that's fine. I'm half wondering whether I should investigate the `abstract` base-type scenario for client, but ... perhaps that is already served by the `interface` API. Either way, there's no need to drag those thoughts into here.

So: have I understood that correctly?

mythz:
Yep that's the server implementation. 

It's more useful when you want to dynamically generate GRPC APIs like exposing existing modules like ServiceStack Services as GRPC APIs, but honestly I'd prefer to create GRPC Services by only sending messages (i.e. w/o interfaces), IMO using interfaces creates unnecessary client/server coupling, friction and inhibits genericizing higher-level functionality as all Unary Messages can go through a single `SendAsync<TResponse>(request)` API which you can easily intercept and apply common functionality to, e.g. ServiceStack Service clients automatically inject SessionId's and BearerTokens to all Request DTOs that implement common interfaces and stuff like auto-retry / with new JWT Token fetched from Refresh Token (e.g. all C# ServiceStack Clients support) is much easier to implement when all messages go through a generic API.

Either base class or marker interface is ok, benefit of interface ofc is you can apply it to existing classes that have their own base class.

mgravell:
OK; moving forward, then; what's your thoughts on `IGrpcService` => `[Something]` ?

I'm thinking in particular of the cost of `serviceType.GetInterfaces().Any(x => x == typeof(IGrpcService))` vs `Attribute.IsDefined(serviceType, typeof(SomethingAttribute))`

mythz:
Either can work, although I prefer the marker interface as it's a better indicator that the primary purpose of this class is for GRPC Services, it's also easier to find all classes implementing an interface (i.e. R#/Rider navigation tools). Also you can add extension methods to interfaces (and default method interfaces) which will allow you to expose common functionality that all `IGrpcService` can benefit from.

GoNextGroup:
Will this approach be added next release?

mgravell:
@GoNextGroup  probably

@mythz  looking at the reference source, I think `typeof(IGrpcService).IsAssignableFrom(type)` should be fine; any chance you could switch to that?

mythz:
Yep good call, done.

GoNextGroup:
Marc, It would be cool to have the same approach. Now I have my similar realization through `CommandTable` which is a dictionary of pairs "Method name, Method Delegate". And one method `Execute` in interface realization. All methods need to have base request type and response (that really simplifies assembliing and parsing method in the `CommandTable` and calling them. 
This approach really simplifies my code approach - I do not need to create `CommandTable`, do not need to create additional wrapper to the request which is only needed for command name passing and so on.
So, thank you you both!

mgravell:
thanks for that; sorry to keep peppering you with questions, but I want to make sure the API is right here.

Additional thought:

image

Are we sure that `type.Name` is always the right thing here? Can this cause problems mapping to arbitrary gRPC services? Does it force namespaces etc in ways that might not suit? Is there a better way we might want to do this? (it would be trivial on the attribute approach, for example)

mythz:
With this POCO convention-based approach you'd naturally adopt the Services type name. It's easy enough to have it also check for any `[ServiceContract]` but honestly I wouldn't do it unless someone presents a good use-case for having separate Services Type Name vs Service Name. By default we can assume everyone's adopting the default convention, there's also simplicity in enforcing a convention, no-one needs to be concerned about mappings and whether they can be different and where the mapping information is maintained, etc. If someone absolutely needs a different name, e.g. the services type is defined in an assembly they can always sub class it and choose a different name and register that type instead, an option not available with attributes.

It's easy enough to add support for it if needed later, but I'd enforce the convention first unless there's a good reason not to.

mgravell:
service name is probably fine, but *typically* gRPC services are disambiguated by the package too; for example, with [this service](https://github.com/protobuf-net/protobuf-net.Grpc/blob/master/examples/grpc/Shared/TimeService.proto), the actual service name is `"MegaCorp.TimeService"`, not `"TimeService"`. So with that in mind; should this be `FullName` ? or...?

GoNextGroup:
I'd love to give my two cents if you'd like, but it's really not a good idea to add namespace name to the service name, because namespace could be complex, like "MyBest.TheCoolest.BestOfall.MegaCorp.Core.Data" namespace. Probably, in this case namespace were added for dividing TimeiService on the MegaCorp namespace and maybe, same other TimeService, weren't it?

mythz:
IMO using C# namespaces would be fragile where simple innocent refactorings can break clients. I've never leaked C# namespaces in external service contracts, the name of the Type is always the name of the Service.

We can already use URL to create a unique URI to a Service, don't see why you'd need it on the Services type as well. If you want you can provide an optional prefix when registering the Type, but it's not something I'd ever use myself. If it was up to me I wouldn't even have GPRC's separate Service Name, all Request DTO types would map 1:1 with the Service they want to call, again makes a tonne of things simpler by being able to send a Request DTO to an endpoint and have it invoke a service whether by HTTP/MQ, it means all you need is the message type to invoke a service.

mgravell:
Fair enough;  suggest we formalize this in `IGrpcService.cs`, i.e adding to the intellisense:

> When a type implements this interface, the type name (without any prefix or other qualification) will be used as the name for resolving the gRPC service. All public methods defined by the type will be mapped as gRPC methods, using their declared name.

With that (assuming the above summary is correct?): I think we're all good.

mgravell:
Thanks; I'll get you a nuget when I'm at my desk, so you can play with it more.

mythz:
Cheers 👍, already using an interned version with these modifications, will be nice to ditch my fork and reference the official package instead.

GoNextGroup:
> With this POCO convention-based approach you'd naturally adopt the Services type name. It's easy enough to have it also check for any `[ServiceContract]` but honestly I wouldn't do it unless someone presents a good use-case for having separate Services Type Name vs Service Name. By default we can assume everyone's adopting the default convention, there's also simplicity in enforcing a convention, no-one needs to be concerned about mappings and whether they can be different and where the mapping information is maintained, etc. If someone absolutely needs a different name, e.g. the services type is defined in an assembly they can always sub class it and choose a different name and register that type instead, an option not available with attributes.
> 
> It's easy enough to add support for it if needed later, but I'd enforce the convention first unless there's a good reason not to - it's the way of the POCO :)

If it really needs it could be implemented through the Interceptor which could be added on the server side, taking the request, looking the method calling name and calling the masked method (method masking could be realized through the method attribute like [Masking(OriginalMethod, PublicNameMethod)].

mgravell:
1.0.10; thanks

PR #28, opened 2019-10-16
Spin up an AppVeyor build

NickCraver:
Once this is merged we should be able to add in Appveyor and be good to go. Tested and working locally, since this one's simpler in directory structure, it's a bit easier to reason about and no shenanigans.

The file moves are just moving `Play*` to `/toys/` instead of `/tests/` to make it simpler (matches what we did in SE.Redis).


NickCraver:
Gonna get this going ASAP

Issue #27, opened 2019-10-16
Use struct as parameters and return value.


Ilchert:
I am trying to create simple method with `int` parameter and it fails with `System.Security.VerificationException: 'Method ProtoBuf.Grpc.Configuration.ServerBinder.AddMethod: type argument 'System.Int32' violates the constraint of type parameter 'TRequest'.'` is it expected behavior? Should I wrap it to some request class? Are there any limitations to use struct as parameters? Will it be allowed in the feature?

mgravell:
Yes, I'm aware of the struct limitation, and yes the library should hide it from you automatically. So: yes, it is a bug and yes, I'll fix it.

This is a requirement in the underlying gRPC API, but there's no need for me not to hide it.

It was actually already on my list because protobuf-net v3 adds value-type implementations of `Empty`, `Duration`, `Timestamp`, etc. But: great to have it listed here to make sure I get it done.

mgravell:
Just as an implementation note: what it'll actually do here will depend on the type; if the value-type being sent is a scalar, it'll treat it as though it were a message with a single field with field number `1`, of the supplied type and value; if the value-type being sent is a message, it'll send the value *as the root message*, but in a way that makes gRPC happy.

Ilchert:
Nice, thanks for the clarification!

Sossenbinder:
I was just wondering about the progress on this issue - Did you manage to find time to do this as of yet?

nilmas:
This enhancement would be very welcome.

mgravell:
I'm wondering if we can fix this at the core API instead ^^^

LMSDev:
@mgravell I recently came across your library (fantastic work, btw!) and encountered this type constraint. Additionally, I found gRPC's lack of support for multiple parameters discouraging, especially for simple method signatures. I understand the reasoning behind it, but it would still be nice to use them. 

As a PoC I wrote a dynamic proxy that intercepts all calls with more than 1 argument (and where the second argument is not `CancellationToken`) and generates a new type with fields based on the arguments. This type is then serialized to `byte[]` using protobuf-net and passed to a more generic gRPC call along with the function name. The server uses the function name to locate the function signature and is able to also generate a type suitable for deserialization. The field values are then extracted from the deserialized type and passed to the original function (with multiple parameters), which happily does its work. Returning a value works exactly the same way.

[...]

PR #30, opened 2019-10-17
Update Reshape.cs

maniglia:
Method ProtoBuf.Grpc.Internal.Reshape.UnaryValueTaskAsync: type argument 'System.Boolean' violates the constraint of type parameter 'TResponse'


mgravell:
I *think* what you're trying to do here is covered by #27? if so: yes, it is on the list.

It is a bit more complex than this, though - I need to spoof around some internal demands that the underlying API places on all callers.

PR #32, opened 2019-10-18
light up the protobuf-net capability API, for array-segment, buffer writer, and ROS

mgravell:
protobuf-net 2.4.1 adds a new capability API, and implements it for array-segments; use that

protobuf-net 3.* implements that capability API for buffer-writer/ROS too; detect and use that


JamesNK:
FYI 2.25.0-pre1 is on NuGet and supports buffer writer.

mgravell:
superceded by https://github.com/protobuf-net/protobuf-net.Grpc/pull/53 because case sensitivity broke git

Issue #33, opened 2019-10-25
Make Callcontext injectable


ElderJames:
Can I inject the `Callcontext` to the constructor of service implement？
Just like the `IHttpContextAccssor` that it does not need to be a method parameter.

mgravell:
No; the call-context is strictly per operation - it has a very different lifetime than the service (on the native version, the service is essentially a singleton; on Kestrel, the service lifetime is configurable in many different ways). I'd also rather avoid some kind of ambient thread-local API.

Can you clarify: what is the scenario where you need it, but are reluctant to make it a parameter?

ElderJames:
I want to get data from headers of the current `HttpContext`,like `jwt`,better to make the `IHttpContextAccssor` work in the service on Kestrel.

mgravell:
OK; I understand what you mean.

@JamesNK: before I go too far down the rabbit-hole investigating this - is there any ambient gRPC context already implemented in your bits?

ElderJames:
And I also need to get the complate signal from client at duplex streaming mode

mgravell:
@ElderJames that sounds like a very different thing, but that *should* currently materialize as the end of the `IAsyncEnumerable<T>` sequence. Is that now working for you? Or are you after something different? Perhaps an example of your intent would help there

ElderJames:
Ok,  `yield break`  for IAsyncEnumerable<T> is work.

mgravell:
I'm assuming that `HttpContextAccessorExtensions.GetServerCallContext` / `.GetCallContext` do what is needed here

PR #35, opened 2019-10-27
add HttpContextAccessor to surppot server context injecting (#33)

ElderJames:


NickCraver:
Having the new extensions is nice, but why does this need to always inject one? Remember that there is an async cost (that’s why it’s not added by default) - I’d argue very few things should inject the context accessor, only doing so if it’s critical to their functionality.

As an aside: the namespace change for the existing extension is a good one, that’s the recommendation so intellisense works better.

ElderJames:
I thinks we can implement the EnableCallContextPropagation optional method like server-side cilent in Grpc.Net, the difference here is code first.
We need to propagat the data likes authentication, trace, configration etc. with httpcontext headers during cross-service invocation.

NickCraver:
I'm not sure how the changes are related though, in this PR. The extension methods work on an HttpContextAccessor, which you can get if one is registered. The other change is to _always register one_, regardless of if it's used. I strongly recommend not doing this in any library that doesn't absolutely need it. Docs telling the user to register one if they want to use it would be the appropriate path there. This is the common pattern, and why it has a specific extension method built into the framework to make it easy to do.

ElderJames:
Ok, I have reverted it.

mgravell:
Why are you changing the na.espace of an existing type? That's a huge breaking change. Also, why are you declaring things in `Microsoft.AspNetCore.Http`? Are there reasons? This feels wrong...

NickCraver:
Oh hmm didn't realize this was 1.x - FWIW the namespace change to `.DepdendencyInjection` is correct but yeah: breaking. Microsoft has official guidance here: https://docs.microsoft.com/en-us/aspnet/core/fundamentals/dependency-injection?view=aspnetcore-3.0, specifically:

> We recommended that apps follow this convention. Place extension methods in the **Microsoft.Extensions.DependencyInjection** namespace to encapsulate groups of service registrations.

I'm not sure about patterns w.r.t. HttpContextAccessor - agree on not putting there unless specifically instructed to do so.

mgravell:
If moving it is correct: fine - but what we need to do is:

- leave the old type/methods in place (have them call the new ones)
- remove the `this` so they aren't extension methods any longer
- add `[Obsolete]`

By doing so, it isn't a hard break.

mgravell:
Note: the "new" declarations only needs one method - the overload was because the v1.0 lacked the return value and second parameter. We don't need the overload in the new code - just make it one method with an optional parameter

mgravell:
Looking at the guidance and context here, my vote would be to move both the `AddCodeFirstGrpc` and call-context extension methods to:

`Microsoft.Extensions.DependencyInjection.CodeFirstGrpcExtensions`

Thoughts?

ElderJames:
Sorry,I think we can just add a call-context extension class in this PR and leave other changes to another.
The reason to declare it in `Microsoft.AspNetCore.Http`  is simply to follow the namespace which it extends.

PR #45, opened 2019-12-05
Added interceptor support to AddCodeFirst

Donis-:


mgravell:
Appreciate the work here - was definitely a gap.

I'm fine with the concept here, but I'm going to merge it as-is, and then tweak a few things afterwards to help it - in particular:

1. make it not be a breaking API change (method forwarding, etc)
2. just use `IEnumerable<T>` - no need for a list API

Thanks

mgravell:
See also #7220cc879363fb234874e674dd24c9832c244844

Issue #38, opened 2019-11-04
Allow OperationContract with multiple arguments


lanfeust69:
I'm trying to evaluate a replacement of WCF with gRPC, within a big code base with interfaces as "code first" approach. So your work here is very highly appreciated !

So far, I've had decent results, but aside from #27, the biggest challenge for using the exact same interfaces is that we often have operations that have multiple arguments. I can see a number of workarounds, but I was wondering whether it could make sense to handle this case inside protobuf-net.Grpc, basically stuffing all arguments (except a last `CallContext`) in a Tuple (or Tuple-like). I'll try to give it a shot, but would be interested in your thoughts about it.

mgravell:
Yup; agree there is scope to improve this, but my main internal question is : what happens when you change the number of parameters either *from* one, or *back* to one, as changing from:

```
message Foo {...}
```

to

```
message InvisibleWrapper {
    Foo foo = 1;
    Bar bar = 2;
}
message Foo {}

lanfeust69:
Here is what I'm trying (and so far it works well on my use-cases) : in [ContractOperation](https://github.com/protobuf-net/protobuf-net.Grpc/blob/master/src/protobuf-net.Grpc/Internal/ContractOperation.cs#L173), if there is more that one argument of `TypeCategory.Data`, I jam all of these into a single one, a `Tuple<>` with appropriate elements. I hope I'll be able to show something more concrete, but although our company officially encourages contributing to OSS, there are still a few hoops to jump through...

Regarding the breaking changes, I'm not sure I get your point : isn't it the same kind of problem if you change the type of the single Request object of an operation ?

mgravell:
> isn't it the same kind of problem if you change the type of the single Request object of an operation ?

perhaps; and perhaps it is a non-issue, that comes under the heading of "yeah, don't add/remove parameters and expect nothing to break". So; perhaps literally doing nothing except wrapping it is fine.

mtsfoni:
I am in the same place @lanfeust69 was 4 years ago (Yeah, we still run WCF and .net framework in 2024...).

I wonder, did you find a solution for this and #27 that you can share? I want to be able to make an interface in whatever way we deem it right, and use this interface in the client. So "Business-Logic developers" do not need to consider the quirks of the network technology or the related package (When building a prototype for our future stack, I just stumbled over getting an notSupportedException when I have two parameters). Also, I believe true code-first should not force the developer into creating the interface in a specific way.

[...]

PR #48, opened 2019-12-13
Improve compatibility with legacy service contracts

lanfeust69:
A great use of the code-first approach offered by `protobuf-net.Grpc` is the ability to smoothly migrate from a WCF-based service architecture to a gRPC one, with as less code change as possible.

There are currently two major incompatibilities when trying to use existing `ServiceContract` interfaces : handling of operations with multiple parameters (#38), and of value types (especially non-string primitive types such as numeric types, and structs such as DateTime), see #27.

This PR proposes to handle these cases by wrapping things into `System.Tuple`s when appropriate.

### DIFF
--- a/src/protobuf-net.Grpc/Configuration/ServerBinder.cs
+++ b/src/protobuf-net.Grpc/Configuration/ServerBinder.cs
@@ -80,8 +80,10 @@ namespace ProtoBuf.Grpc.Configuration
 
                     if (argsBuffer is null)
                     {
+                        var marshallerFactory = new ValueTypeWrapperMarshallerFactory(binderConfiguration!.MarshallerCache);
+                        var marshallerCache = new MarshallerCache(new[] { marshallerFactory });
                         argsBuffer = new object?[9];
-                        argsBuffer[6] = binderConfiguration!.MarshallerCache;
+                        argsBuffer[6] = marshallerCache;
                         argsBuffer[7] = service is null ? null : Expression.Constant(service, serviceType);
                     }
                     argsBuffer[0] = serviceName;
--- a/src/protobuf-net.Grpc/Configuration/ServerBinder.cs
+++ b/src/protobuf-net.Grpc/Configuration/ServerBinder.cs
@@ -163,7 +163,7 @@ namespace ProtoBuf.Grpc.Configuration
                     else
                     {
                         // basic - direct call
-                        return (TDelegate)Delegate.CreateDelegate(typeof(TDelegate), _service, Method);
+                        return (TDelegate)Delegate.CreateDelegate(typeof(TDelegate), _service?.Value, Method);
                     }
                 }
                 else
--- a/src/protobuf-net.Grpc/Internal/ContractOperation.cs
+++ b/src/protobuf-net.Grpc/Internal/ContractOperation.cs
@@ -221,7 +221,7 @@ namespace ProtoBuf.Grpc.Internal
 
             if (!s_signaturePatterns.TryGetValue(signature, out var config)) return false;
 
-            static Type GetTupledType(Type[] types)
+            static Type GetTupledType(params Type[] types)
             {
                 return types.Length switch
                 {
@@ -238,8 +238,21 @@ namespace ProtoBuf.Grpc.Internal
                 };
             }
 
-            Type? arg0Type = dataArgsTypes.Length > 1 ? GetTupledType(dataArgsTypes) :
-                args.Length > 0 ? args[0].ParameterType : null;
+            Type? arg0Type = null;
+            if (dataArgsTypes.Length > 1)
+                arg0Type = GetTupledType(dataArgsTypes);
+            else if ((dataArgsTypes.Length == 1 && dataArgsTypes[0].IsValueType))
+                arg0Type = typeof(ValueTypeWrapper<>).MakeGenericType(dataArgsTypes[0]);
+            else if (args.Length > 0)
+                arg0Type = args[0].ParameterType;
+
+            Type retType = method.ReturnType;
+            if (signature.Ret == TypeCategory.Data && retType.IsValueType)
+                retType = typeof(ValueTypeWrapper<>).MakeGenericType(retType);
+            else if (signature.Ret == TypeCategory.TypedTask && retType.GetGenericArguments()[0].IsValueType)
+                retType = typeof(Task<>).MakeGenericType(typeof(ValueTypeWrapper<>).MakeGenericType(retType.GetGenericArguments()[0]));
+            else if (signature.Ret == TypeCategory.TypedValueTask && retType.GetGenericArguments()[0].IsValueType)
+                retType = typeof(ValueTask<>).MakeGenericType(typeof(ValueTypeWrapper<>).MakeGenericType(retType.GetGenericArguments()[0]));
 
             (Type type, TypeCategory category) GetTypeByIndex(int index)
             {
@@ -248,7 +261,7 @@ namespace ProtoBuf.Grpc.Internal
                     0 => (arg0Type!, signature.Arg0),
                     1 => (args[1].ParameterType, signature.Arg1),
                     2 => (args[2].ParameterType, signature.Arg2),
-                    RET => (method.ReturnType, signature.Ret),
+                    RET => (retType, signature.Ret),
                     VOID => (typeof(void), TypeCategory.Void),
                     _ => throw new IndexOutOfRangeException(nameof(index)),
                 };
--- a/src/protobuf-net.Grpc/Internal/ContractOperation.cs
+++ b/src/protobuf-net.Grpc/Internal/ContractOperation.cs
@@ -205,21 +205,47 @@ namespace ProtoBuf.Grpc.Internal
             if (method.IsGenericMethodDefinition) return false; // can't work with <T> methods
 
             if ((method.Attributes & (MethodAttributes.SpecialName)) != 0) return false; // some kind of accessor etc
-            
+
             if (!binderConfig.Binder.IsOperationContract(method, out var opName)) return false;
 
             var args = method.GetParameters();
-            if (args.Length > 3) return false; // too many parameters
+            var dataArgsTypes = args
+                .TakeWhile(arg => GetCategory(binderConfig.MarshallerCache, arg.ParameterType, bindContext) == TypeCategory.Data)
+                .Select(arg => arg.ParameterType)
+                .ToArray();
+            if (args.Length - dataArgsTypes.Length > 3) return false; // too many parameters
+            if (dataArgsTypes.Length > 1)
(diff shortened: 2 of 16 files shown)


mgravell:
Haven't reviewed yet, sounds intriguing, but: question (again, I haven't looked at the code yet); how are value-type message types treated currently? Would they get wrapped? If so, that is undesirable. I may have a way around that...

lanfeust69:
For now I indeed wrapped value-types. My idea was that it's not worse than other forms of boxing, especially when there is network involved anyway, which puts the cost of allocation in perspective. And it seemed rather logical, carrying on what I did for multiple parameters, with the same zones of code impacted.
Still, "boxing" is indeed undesirable, and I admit that it was a bit more complex than I initially thought, mainly the code generation to transform a `ValueTask<Tuple<T>>` into a `ValueTask<T>`. I kept that in a separate commit though, it should be easy to try something else. I'm not sure how much time I can find, but I may explore other venues if you tell me what alternative(s) you had in mind.

mgravell:
The issue here isn't one of allocations. Totally agree that in the grand scheme of things, that isn't going to matter. What I'm trying to preserve is message semantics. So: if you send a `class Foo` with some contract members, and later change it to `struct Foo` with the same members: it should continue to work identically. I suspect we can do this trivially in protobuf-net V3 via a generic wrapper type with a custom serializer, using WriteAny etc internally. However, those details don't matter right now. What matters is that we should design for that end game.

So: if the current code in this PR would make that impossible, I would say that it would actually be better for the code to detect this case and throw a not-supported-exception, rather that end up deploying a wrapped message which would be incompatible with the aim. Then we can fill in the gap. If we ship it in the "yay, this works" way, it is **really** hard to change it later, since it breaks comms that may then exist.

Do you see what I'm trying to prevent?

lanfeust69:
Yes, I see your point. There are things that are part of the communication contract, and struct/class isn't one of them. So a contract-preserving change should not break things, and what goes over the wire should only depend on the contract.

I (of course) have been influenced by my own use-case, where clients and servers are regularly synchronized, so that any breaking change in the communication mechanism is not such a big deal, we simply schedule these for major releases. And I think we never changed from a struct to a class (or the other way) 😃 .

I'll try to find some time to think things over and experiment...

lanfeust69:
I changed a bit the implementation to handle value types : there is still a wrapping (I think it is necessary for proper interop with Grpc.Core), but in a specific ValueTypeWrapper class, which allows to include a marshalling step that recognizes it and actually marshalls the underlying type, preserving any contract semantics.
I'm not sure about primitive types, though : currently I don't treat them differently, so that the payload over the wire is just the serialization of the parameter itself, therefore not within a message, which is not compliant with a proto3 rpc definition (if I read the specs correctly). I guess it wouldn't be difficult to handle this, mostly a matter of not missing "semi-primitive" types such as `DateTime` or `TimeSpan`, but I'd rather have your feedback first.

mgravell:
K; this is on my radar to look at, but note that I'm on limited
availability over the xmas holidays.

lanfeust69:
> note that I'm on limited availability over the xmas holidays.

Sure ! Actually, I had a bit of time mainly because we've had almost 1 meter of snow falls during the last three days, so I spent much less time outside than I would have, had the weather allowed it.

MarcDrexler:
Hi @lanfeust69 

I have just seen this conversation and am very happy that something has already been experimented with multiple parameters 😄 
We have a similar case where clients and servers are under our control, and major changes to the contract are not a showstopper for us.

But now I wonder if there is any further progress here. Is there still something missing, or is this code already used by you today?

lanfeust69:
Hi @MarcDrexler 

I have been working on some other subjects for a while, so left it as it was : no known issue from the few proof-of-concept tests I did (with real production WCF contracts though). But I think I'll be back on this soon. Of course I'd rather use this code through a released package, but this is not mandatory for us.

@mgravell : did you find some time to have a look at the PR ?

lanfeust69:
I pushed a slightly different version : need for wrapping is done using `Type.IsValueType` instead of `!Type.IsClass`, which is clearer, and handles the situation of an interface (typically `IList<Something>`) in an operation signature.

Will look into the merge conflict later...

zhech2:
We are very interested in the changes. They would help a lot with our current project. Is there anything we could do to help get this or a similar change completed?

mgravell:
Mostly I need to find the time to go back through it. I've also proposed PRs to the core gRPC repos to try and light up the value-type thing properly, so of particular concern is that any value-type single parameter handling needs to be byte-for-byte identical with what the marshaller would do without any wrapping, so that we can potentially remove that code eventually.

zhech2:
I built protobuf-net.Grpc with this pull request and used it in my project.  It worked in all the scenarios I had in moving a WCF service to GRPC.

geoffreygeslin:
Like previous posts, I'm really interested on this request...
In short, my company is in a similar case
- Had a huge codebase with MANY Wcf DataContracts and ServiceContracts ;
- Clients and servers are under our control and updated simultaneously;
- We want to migrate to .Net Core + gRPC,  without editing former contracts interfaces.

We'll greatly appreciate to see the PRs integrated to main repo.

zgabi:
I'm also using this PR in our application. It works. (We have the server and client side, too, so don't care about any future change or incompatibility with newer releases)

I updated it to the latest code: https://github.com/zgabi/protobuf-net.Grpc
And created a a nuget package: 
https://www.nuget.org/packages/ZGabi.protobuf-net.Grpc/
https://www.nuget.org/packages/ZGabi.protobuf-net.Grpc.Native/
https://www.nuget.org/packages/ZGabi.protobuf-net.Grpc.AspNetCore/

I hope this will be implemented in the official package.

mgravell:
I fully intend to go through it and consider a number of problematic edge cases; it is simply a capacity (my capacity) problem. I've had a *ton* of other things to work on, and ultimately this is an unpaid, unsponsored side project that needs to fit in alongside my other commitments.

I'm actually kinda glad we've held back though - through initial usage, a number of those edge cases have come to mind. Additionally, V3 of protobuf-net has the new "compatibility level" feature that impacts a few primitives, and I'll probably try to make sure that this gets reflected appropriately here too, i.e. so that a `Guid` parameter gets serialized as a `string` when the compatibility level is 300 or above, to improve cross-platform friendliness.

owaits:
I would also like to add my weight to seeing this in the official package. When you are migrating from a WCF contract you really don't want to have to re-write all your methods. I understand in the future you probably want to avoid writing methods in this way but if your moving code it makes sense to be able to support this case. 

For the moment I am using the nuget package from @zgabi which works really well. Thank you so much to everyone for there work on this, completely understand time pressures etc. Hopefully @mgravell you can find time to look at this, is there anything the I can do to help you with this?

Gambero81:
This is a great pr!

zgabi:
I've updated the packages to version 1.0.136.

Gambero81:
Hi @mgravell i know you are full of work, have any plan to accept this pull request?
It's a "must have" for converting wpf projects to grpc, or to simplify the usage of the library.

Thanks

mgravell:
Ultimately "yes", but it is a complex set of changes that needs some
dedicated time (and I think some of the scenarios are now even more
complicated due to parallel code changes that have added an entire other
dimension to what needs to be considered, in terms of possible call
signatures).

Another part of me is wondering whether we're trying to solve this the
wrong way; I'm kinda tempted to see whether we can also crack this nut with
"generators", meaning: for client side we could emit the entire proxy *at
build*, including a dummy type for wrapping the args, and do something
similar at the server. But: that's just another thought echoing around my
mind.

Right now I'm trying to ship some Dapper.AOT pieces, before I can get back
onto this.

Gambero81:
@mgravell take a look to this project to handle multiple parameter, value type and source generators:

https://github.com/max-ieremenko/ServiceModel.Grpc

fredliex:
Is there any progress?
I have been waiting for this PR....

fredliex:
@zgabi . Have you considered upgrading `ZGabi.protobuf-net.Grpc` to 1.0.147?

The waiting process is always long.

fredliex:
hi @lanfeust69 

The following error occurs when calling `string SomeMethod(int? value)`.

```
System.ArgumentException
  HResult=0x80070057
  Message=GenericArguments[0], 'System.Nullable`1[System.Int32]', on 'ProtoBuf.Grpc.Internal.ValueTypeWrapper`1[T]' violates the constraint of type 'T'.
  Source=System.Private.CoreLib
  StackTrace: 
   at System.RuntimeType.ValidateGenericArguments(MemberInfo definition, RuntimeType[] genericArguments, Exception e)
   at System.RuntimeType.MakeGenericType(Type[] instantiation)
   at ProtoBuf.Grpc.Internal.ContractOperation.TryIdentifySignature(MethodInfo method, BinderConfiguration binderConfig, ContractOperation& operation, IBindContext bindContext) 
```

```c#
else if ((dataArgsTypes.Length == 1 && dataArgsTypes[0].IsValueType))
   arg0Type = typeof(ValueTypeWrapper<>).MakeGenericType(dataArgsTypes[0]);  // <-- exception here
```

The struct generic constraint cannot accept `Nullable<T>`. 
ref https://docs.microsoft.com/en-us/dotnet/csharp/programming-guide/generics/constraints-on-type-parameters
```
public class ValueTypeWrapper<T> where T : struct
```

I tried to remove the following three struct generic constraint, and it worked as expected.
```
public class ValueTypeWrapper<T> //where T : struct
public class ValueTypeWrapperMarshallerFactory {
   private void InvokeSerializer<T>(......) //where T : struct
   private ValueTypeWrapper<T> InvokeDeserializer<T>(......) //where T : struct
}
```

The above let you know.

lanfeust69:
@fredliex : indeed, I came across this one, it has been a bit of a surprise that `Nullable<>` is a `struct`, except it is not 😃 .

I dealt with it in a local fork (along with a few other edge cases such as parameters passed by `ref`/`out`), but there are also more specific things, and I haven't had time to sort out what should be added to this PR, what should be in a separate one, and what is not of general interest.

zgabi:
@fredliex upgraded to the latest code.
Also added you fix for the nullable types (removed the constaraint)

menaheme:
Any news on this?

czaccardelli:
Would be nice to see this (or this functionality in any form) incorporated not just for migrating WCF service contracts easier but simply to avoid the awkward wrappers around primitives and also method signatures requiring complex objects as parameters.

mgravell:
I'm going to close this one; I really appreciate the input, but when combined with some other changes that have happened in the binder, it is clear that this is going to cause a complexity problem in the binder. What I'd like to do instead is leave the library alone here, and implement this feature via the AOT approach. The main pieces for this are in progress separately.

Again, huge thanks, but I don't think this is the way we should go here.

PR #85, opened 2020-04-30
add first stab at generators; doesn't work well yet

mgravell:
Current status: package ref of protobuf-net.Reflection breaks the world


gustav3d:
You dont want to use https://devblogs.microsoft.com/dotnet/introducing-c-source-generators/  ?

mgravell:
@gustav3d if you pinged me on this (I'm a little confused by an email notification): that's exactly what it is using

gustav3d:
Yes, your correct !  i got confused by that there was .reflection as name.
What your doing is great work :)

Myself is looking to somehow be able to from a server write pre serialized
entities that i have cached in RAM, (that i auto update on changes, updates
are rare vs read requests) .

You got any idea how i can implement that in your APi?.

Thanks
  Gustav

mgravell:
@gustav3d I can't decipher the scenario you are describing. Probably the best thing would be to illustrate it with an example of what you are trying to do, if an imaginary yet somehow perfect API existed.

Issue #14, opened 2019-08-27
streaming an IObservable


apobekiaris:
using grpc with a proto file as per asp.net docs i wrote
```
service Dispatcher {

	rpc RequestEvent (Request) returns (stream Event) {}

```

the grpc tools generate the service which I then integrate to the existing observables in my system like  

```cs
 internal class DispatcherImpl : Dispatcher.Dispatcher.DispatcherBase{
        
        public override async Task RequestEvent(Request request, IServerStreamWriter<Event> responseStream, ServerCallContext context){
  
                
                await SomeController.Observable
                    .ForEachAsync(value =>
                        {
                            responseStream.WriteAsync(value).Wait();
                        }, 
                        context.CancellationToken);


        }
```
this pretty much works ok, but I would like to send .net types and share them and avoid the `build hack` so I came here. However all your samples use the IAsyncEnumerable, although you mentioned that Task is also possible. So, can you provide an example based on your `TimeService` perhaps that streams an IObservable?

mgravell:
> However all your samples use the IAsyncEnumerable,

Right now, `IAsyncEnumerable` is how protobuf-net.Grpc sees `stream`; additional APIs may be possible in theory, but that's the primary API anticipated

> although you mentioned that Task is also possible

Can you be more specific as to the context there? I can't give you a `[Value]Task<List<T>>`, for example - the entire point of a `stream` is that it is an *open* sequence, and blocking an open sequence is anathema. You mention `IObservable` - "reactive" is indeed a very similar and alternative model; it *isn't* once I've touched so far, but I could *absolutely* be agreeable to that shape of API; however, isn't the point of IObservable that it is *similar to a sequence* (push instead of pull), so surely it would be a *single* observable, not a sequence?

apobekiaris:
>Can you be more specific as to the context there? 

quote from docs: We’re using ValueTask<T> here, but the library also supports Task<T>

```cs
await SomeController.Observable
                    .ForEachAsync(value =>
                        {
                            responseStream.WriteAsync(value).Wait();
                        }, 
                        context.CancellationToken);
````

what happens above while we are on traditional .net grpc is: I await for an observable that never completes and for each emit i send to the wire.

>Right now, IAsyncEnumerable is how protobuf-net.Grpc sees stream

how does this happen? can i directly write to the stream as above? having a Task as a return type?

mgravell:
> quote from docs: We’re using ValueTask here, but the library also supports Task

[...]

PR #92, opened 2020-06-03
explore "push enumerable" ideas from #14

mgravell:
This is a tracking branch to explore ideas around push async enumerables


JuliusSweetland:
@mgravell I'm just experimenting with this implementation in my code. If it makes it into prod can you add support for completeWriterOnCancellation (on AsAsyncEnumerable method)? There's no way to achieve this at the moment as those methods expect a Channel<T>

mgravell:
@JuliusSweetland only just noticed your comment; will look in the week - thx for the feedback

mgravell:
I'm not sure I'm happy with this; killing for now

JuliusSweetland:
@mgravell I liked the idea, but I didn't need this implementation in the end. Thanks though.

PR #108, opened 2020-07-05
Add SchemaGenerator API

mgravell:


gustav3d:
Request and response has replaced each other in all the generated rpc methods.

mgravell:
Yep, you're right. Will fix.

mgravell:
@gustav3d fixed in protobuf-net 3.0.23

gustav3d:
Thanks  a lot for the quick fix !!

gustav3d:
For Stream, It's on the request param instead of the return.

gustav3d:
For server streaming that is.

mgravell:
I had one job! The sad thing is, I genuinely wrote a test for this. I just screwed up the test in exactly the same ways as the code. Thanks, will rectify. Again.

(Quietly looks embarrassed)

mgravell:
3.0.24 @gustav3d

gustav3d:
Thanks for fixing so fast !
It seems to work well now :)

JohnGalt1717:
Is there any documentation on how to use this to, say put up Swagger GRPC or create a simple file browser within your service that protoc can use to generate say a dart client from the results?

JohnGalt1717:
I run this on this on the following service defintion and I get: Data of this type has inbuilt behaviour, and cannot be added to a model in this way: System.String
```
	[ServiceContract(Name = "SystemService")]
	public interface ISystemService : IService
	{
		ValueTask<string> Ping(string name);
		ValueTask<string> PingLoggedIn();
	}
```

mgravell:
Indeed, that probably shouldn't work. If it currently does: that's sort of
a bug. In gRPC you're meant to send DTOs, not primitives.

We want to make protobuf-net.Grpc fake over that bit automatically for you,
but that is not currently implemented.

JohnGalt1717:
@mgravell Ok, makes sense and I'm not sure it should be implemented. Might want to clearly document this for protobuf.net though.

PR #119, opened 2020-07-29
[Docs] Forward CancellationToken to Delay call

rwkarg:


rwkarg:
Small update to the example. Just allowing the Task.Delay to cancel early if the CancellationToken is cancelled.

mgravell:
Fair enough; it shows the available of cancellation in a nice way, thanks

PR #129, opened 2020-09-18
fixes the proto definition generation for generic types.

Lutti1988:
methods with generic types in input/output were ignored.


mgravell:
Is there an example / test of when this is useful?

mgravell:
(also: sorry for delay; 2020 happened)

mgravell:
Closing due to a: lack of example/test, but more importantly b: I think this is now more typically handled via `[SubService]` and interface composition

Issue #121, opened 2020-08-07
Issue sharing a gRPC service with DI.


Euan-McVie:
Related to #120 (I've found why we were mapping the interface)
The reason was to allow inter service communication to not have to worry about it being a local or remote service and so we would be free to move services between hosts as required.

In order to do this we register the services with DI in the `startup.cs`, rather than have the gRPC service being a wrapper around another DI registered service. This then means that for the gRPC mapping to use the service we found that we had to use `endpoints.MapGrpcService<ICalculator>();` so that the instance of ICalculator came from DI rather than it's own instance.

This works, but then means that the `[Authorize]` issue is back as although the instance is pulled from DI, the metadata is only pulled from the type that is mapped.

As long as the services are scoped/transient it isn't a big issue that they use a different instance, it is only if the service is a singleton that the problem becomes really apparent.

I have uploaded an example to my [fork](https://github.com/Euan-McVie/protobuf-net.Grpc/commit/67419223efa4fc90961f84f482605e68bc6f2b3b) by extending your *_CS example.
- I extended the `MyCalculator` to store something within the instance.
- I updated `MyTimeService` to use it for the delay as an example of direct inter service use.
- In `Startup.cs` the `MyCalculator` has been registered as a singleton.
- By switching between the endpoint mappings you can see the behaviour switch:
  - Mapping to `ICalculator` works and both the client and server report the delay from the `MyCalculator` singleton.
  - Mapping to `MyCalculator` doesn't work as the client now gets a new `MyCalculator` instance per request.
- If you uncomment the `[Authorize]` on `MyCalculator`:
  - Mapping to `ICalculator` is NOT rejected when the client calls it.
  - Mapping to `MyCalculator` is rejected when the client calls it.

One of the great things about your lib was the ability for us to have the interfaces for remote and local services registered in DI so that the consumer did not need to worry about if it was local or remote, and how it removed a lot of boiler plate code. While we could create a "gRPC controller" wrapper that shares the interface, and just passes through to the real DI service it is a bit of boilerplate that would be good to avoid if we can.

Any feedback is appreciated.
Thanks
Euan

mgravell:
[...]

PR #130, opened 2020-09-21
Add support for configuring a custom servicebinder using fluent methods (#121)

Euan-McVie:
@mgravell 
As per discussion on #121 i've prepared a PR based on moving the metadata lookup to the service binder.
I've split into 3 commits to make it easier to see the actual implementation vs an update to your example to test the scenario.

Note that I moved the entire `GetMetadata` as calling a separate `GetMetadata(MethodInfo methodInfo)` is too late as the `MethodInfo` for the service type would already be incorrect in my use case.
We could split into seperate `GetContractMetadata`() or `GetServiceMetaData()` overrides, but the method ones need the service type for me to resolve before getting the methodInfo so I didn't think it was worth it if I could resolve the ServiceType before calling the base implementation.

If you want to see it with the existing behaviour simply comment out the `.WithBinderConfiguration` line from the `Startup.cs` and it will then start ignoring the `[Authorize]` attribute in the counter service.

Let me know if you need any changes or want the test example changes dropped from the PR.

Thanks
Euan

### DIFF
--- a/src/protobuf-net.Grpc.AspNetCore/ServicesExtensions.cs
+++ b/src/protobuf-net.Grpc.AspNetCore/ServicesExtensions.cs
@@ -1,11 +1,12 @@
-﻿using Grpc.AspNetCore.Server;
+﻿using System;
+using System.Collections.Generic;
+using Grpc.AspNetCore.Server;
 using Grpc.AspNetCore.Server.Model;
 using Grpc.Core;
 using Microsoft.Extensions.DependencyInjection;
 using Microsoft.Extensions.DependencyInjection.Extensions;
 using Microsoft.Extensions.Logging;
 using ProtoBuf.Grpc.Configuration;
-using System;
 
 namespace ProtoBuf.Grpc.Server
 {
@@ -17,7 +18,7 @@ namespace ProtoBuf.Grpc.Server
         /// <summary>
         /// Registers a provider that can recognize and handle code-first services
         /// </summary>
-        public static void AddCodeFirstGrpc(this IServiceCollection services) => AddCodeFirstGrpc(services, null);
+        public static IGrpcServerBuilder AddCodeFirstGrpc(this IServiceCollection services) => AddCodeFirstGrpc(services, null);
 
         /// <summary>
         /// Registers a provider that can recognize and handle code-first services
@@ -28,6 +29,21 @@ namespace ProtoBuf.Grpc.Server
             services.TryAddEnumerable(ServiceDescriptor.Singleton(typeof(IServiceMethodProvider<>), typeof(CodeFirstServiceMethodProvider<>)));
             services.TryAddSingleton(SimpleRpcExceptionsInterceptor.Instance);
             return builder;
+        }
+
+        public static IGrpcServerBuilder WithBinderConfiguration(this IGrpcServerBuilder builder, Action<BinderOptions> binderOptions)
+        {
+            var options = new BinderOptions();
+            binderOptions(options);
+            builder.Services.TryAddSingleton(BinderConfiguration.Create(options.MarshallerFactories, options.ServiceBinder));
+            return builder;
+        }
+
+        public sealed class BinderOptions
+        {
+            public IList<MarshallerFactory>? MarshallerFactories { get; set; } = null;
+
+            public ServiceBinder? ServiceBinder { get; set; } = null;
         }
 
         private sealed class CodeFirstServiceMethodProvider<TService> : IServiceMethodProvider<TService> where TService : class
@@ -64,7 +80,7 @@ namespace ProtoBuf.Grpc.Server
                 where TResponse : class
             {
                 var metadata = bindContext.GetMetadata(stub.Method);
-                
+
                 var context = (ServiceMethodProviderContext<TService>)bindContext.State;
                 switch (method.Type)
                 {
--- a/examples/pb-net-grpc/Client_CS/Program.cs
+++ b/examples/pb-net-grpc/Client_CS/Program.cs
@@ -1,12 +1,12 @@
-﻿using Grpc.Core;
+﻿using System;
+using System.Threading;
+using System.Threading.Tasks;
+using Grpc.Core;
 using Grpc.Net.Client;
 using MegaCorp;
 using ProtoBuf.Grpc;
 using ProtoBuf.Grpc.Client;
 using Shared_CS;
-using System;
-using System.Threading;
-using System.Threading.Tasks;
 
 namespace Client_CS
 {
@@ -21,6 +21,7 @@ namespace Client_CS
             Console.WriteLine(result.Result); // 48
 
             var clock = http.CreateGrpcService<ITimeService>();
+            var counter = http.CreateGrpcService<ICounter>();
             using var cancel = new CancellationTokenSource(TimeSpan.FromMinutes(1));
             var options = new CallOptions(cancellationToken: cancel.Token);
 
@@ -29,10 +30,14 @@ namespace Client_CS
                 await foreach (var time in clock.SubscribeAsync(new CallContext(options)))
                 {
                     Console.WriteLine($"The time is now: {time.Time}");
+                    var currentInc = await counter.IncrementAsync(new IncrementRequest { Inc = 1 });
+                    Console.WriteLine($"Time recieved {currentInc.Result} times");
                 }
             }
-            catch (RpcException) { }
+            catch (RpcException ex) { Console.WriteLine(ex); }
             catch (OperationCanceledException) { }
+            Console.WriteLine("Press [Enter] to exit");
+            Console.ReadLine();
         }
     }
 }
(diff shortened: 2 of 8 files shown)


Euan-McVie:
Hi @mgravell, 
Any chance of a review for what I need to change in order to get this to an acceptable state?
Thanks

src/protobuf-net.Grpc.AspNetCore/ServicesExtensions.cs
```diff
@@ -17,7 +18,7 @@ public static class ServicesExtensions
         /// <summary>
         /// Registers a provider that can recognize and handle code-first services
         /// </summary>
-        public static void AddCodeFirstGrpc(this IServiceCollection services) => AddCodeFirstGrpc(services, null);
+        public static IGrpcServerBuilder AddCodeFirstGrpc(this IServiceCollection services) => AddCodeFirstGrpc(services, null);
```
mgravell, line 21: note that this represents a hard break for semver purposes; I wonder whether it is possible to do instead:

``` c#
// note no "this", so it is no longer an extension method and will not be selected by new builds
public static void AddCodeFirstGrpc(IServiceCollection services) => AddCodeFirstGrpc(services, null);
public static IGrpcServerBuilder AddCodeFirstGrpc(this IServiceCollection services, Action<BinderOptions> binderOptions = null)
{...}
```

This avoids a hard break but gives you the same results.

src/protobuf-net.Grpc.AspNetCore/ServicesExtensions.cs
```diff
@@ -30,6 +31,21 @@ public static IGrpcServerBuilder AddCodeFirstGrpc(this IServiceCollection servic
             return builder;
         }
 
+        public static IGrpcServerBuilder WithBinderConfiguration(this IGrpcServerBuilder builder, Action<BinderOptions> binderOptions)
+        {
+            var options = new BinderOptions();
```
mgravell, line 36: note that *as I understand it* (I'm not a DI expect), this not the normal pattern for "option" scenarios; not sure whether that matters much - this will probably *work*, although it may have atypical timing in the DI pipe

src/protobuf-net.Grpc.AspNetCore/ServicesExtensions.cs
```diff
@@ -30,6 +31,21 @@ public static IGrpcServerBuilder AddCodeFirstGrpc(this IServiceCollection servic
...
+            binderOptions(options);
+            builder.Services.TryAddSingleton(BinderConfiguration.Create(options.MarshallerFactories, options.ServiceBinder));
+            return builder;
+        }
+
+        public sealed class BinderOptions
+        {
+            public IList<MarshallerFactory>? MarshallerFactories { get; set; } = null;
```
mgravell, line 44: not sure what this is doing or why it exists; PR doesn't mention it; what is the goal here? how is this different to the ambient marshaller setup that can be injected?

src/protobuf-net.Grpc.AspNetCore/ServicesExtensions.cs
```diff
@@ -30,6 +31,21 @@ public static IGrpcServerBuilder AddCodeFirstGrpc(this IServiceCollection servic
...
 
+        public static IGrpcServerBuilder WithBinderConfiguration(this IGrpcServerBuilder builder, Action<BinderOptions> binderOptions)
+        {
+            var options = new BinderOptions();
+            binderOptions(options);
+            builder.Services.TryAddSingleton(BinderConfiguration.Create(options.MarshallerFactories, options.ServiceBinder));
+            return builder;
+        }
+
+        public sealed class BinderOptions
```
mgravell, line 42: the public API should have intellisense; if this didn't raise a warning, that's a bug that I should fix

src/protobuf-net.Grpc.AspNetCore/ServicesExtensions.cs
Euan-McVie: Makes sense. I'll switch to your `AddCodeFirstGrpc` suggestion as it fixes both issues.

src/protobuf-net.Grpc.AspNetCore/ServicesExtensions.cs
Euan-McVie: I was building in `Debug` so it wasn't flagging them locally and I didn't review CI beyond seeing it passed. Maybe `TreatWarningsAsErrors` should/could be turned on?

src/protobuf-net.Grpc.AspNetCore/ServicesExtensions.cs
Euan-McVie: This was just a wrapper as it felt a little odd explicitly listing `null` for the marshallers
i.e.
```cs
config => 
{
  config.ServiceBinder = new MyServiceBinder()
}
```
instead of just passing in
```cs
BinderConfiguration.Create(null, new MyServiceBinder)
```

src/protobuf-net.Grpc.AspNetCore/ServicesExtensions.cs
Euan-McVie: It's not an options pattern. but it is fairly common when the config object passed in is not just an options object to be registered, but is actually used inside to configure how things are registered with DI.
I was going to name it `BinderConfiguration`, but that is what the main one is called.

Euan-McVie: Note it doesn't affect the DI pipeline either as everything is still registered/created explicitly as I invoke the `BinderOptions()` so it is all in place before the serviceprovider is built and requests made of it.

src/protobuf-net.Grpc.AspNetCore/ServicesExtensions.cs
Euan-McVie: What is the preference for when we also need the other override for `GrpcServiceOptions`? It's why I added as an extension instead of an override to `AddCodeFirstGrpc`
I could either add as an extra field to the `BinderOptions`(name change to `CodeFirstGrpcConfiguration`) or would you prefer dual options, Not sure I like adding it as an extra field as it would be added as an `Action<GrpcServiceOptions>` property as the underlying Configure is passed the action to lazy evaluate, which is slightly different to the eager evaluation of the `CodeFirstGrpcConfiguration`
```
//Existing + suggested override
public static void AddCodeFirstGrpc(IServiceCollection services) => AddCodeFirstGrpc(services, null);
public static IGrpcServerBuilder AddCodeFirstGrpc(this IServiceCollection services, Action<GrpcServiceOptions>? configureOptions)
public static IGrpcServerBuilder AddCodeFirstGrpc(this IServiceCollection services, Action<CodeFirstGrpcConfiguration>? grpcConfiguration = null)
// Extra
public static IGrpcServerBuilder AddCodeFirstGrpc(this IServiceCollection services, Action<GrpcServiceOptions>? configureOptions, Action<CodeFirstGrpcConfiguration>? grpcConfiguration)

// Or just switch to non action/option/config style
public static void AddCodeFirstGrpc(IServiceCollection services) => AddCodeFirstGrpc(services, null);
public static IGrpcServerBuilder AddCodeFirstGrpc(this IServiceCollection services, Action<GrpcServiceOptions>? configureOptions)
public static IGrpcServerBuilder AddCodeFirstGrpc(this IServiceCollection services, BinderConfiguration? binderConfiguration = null)
// Add Create override to to allow not setting a marshaller.
services.AddCodeFirstGrpc(BinderConfiguration.Create(new MyServiceBinder));
```

src/protobuf-net.Grpc.AspNetCore/ServicesExtensions.cs
Euan-McVie: Or if you were meaning that I should have been wrapping in a "value factory"
```cs
builder.Services.TryAddSingleton(() => BinderConfiguration.Create(options.MarshallerFactories, options.ServiceBinder));
```
then I agree.

src/protobuf-net.Grpc.AspNetCore/ServicesExtensions.cs
Euan-McVie: Think my preference would be just to get rid of my `BinderOptions`/`CodeFirstGrpcConfiguration`
and go with an extra override to `Create` to remove the null marshaller requirement.

Euan-McVie:
@mgravell - Thanks for the review.
I've commented on a few bits.
Is the general approach for the `ServiceBinder` changes themselves ok. If so I might move the commit to a seperate PR as the fluent stuff is just nice to have so can continue discussions with yourself without holding up the functional change.

Thanks

Euan-McVie:
@mgravell - I still intend to tidy this up a bit further as per some of your comments as I would still prefer a cleaner fluent style of setting a custom `ServiceBinder`, but have moved the `ServiceBinder` changes themselves to a seperate PR #138.

PR #136, opened 2020-10-05
Adjust ServicesExtensions to GrpcClientServiceExtensions

almostcake:
Protobuf-net.Grpc is great, but I have faced with a small problem during migration from `Grpc.Net.ClientFactory` as [ServicesExtensions](https://github.com/almostcake/protobuf-net.Grpc/blob/main/src/protobuf-net.Grpc.ClientFactory/ServicesExtensions.cs) do not implement all the matching methods from [GrpcClientServiceExtensions](https://github.com/grpc/grpc-dotnet/blob/master/src/Grpc.Net.ClientFactory/GrpcClientServiceExtensions.cs), especially the ones with `Action<IServiceProvider, GrpcClientFactoryOptions>` which are required for configuration.
This is a small improvement to proxy those methods.


mgravell:
seems fine, thanks

PR #138, opened 2020-11-10
Add support for defining a custom metadata lookup in the servicebinder (#121)

Euan-McVie:
@mgravell 
Hi,
Usual issue with shifting priorities means I've only now had time to revisit this issue. I have split off the `ServiceBinder` part into this PR so that it can be reviewed separately to the more contentions configuration interface changes in PR #130.

Can this be reviewed independently. 
Thanks
Euan

### DIFF
--- a/examples/pb-net-grpc/Client_CS/Program.cs
+++ b/examples/pb-net-grpc/Client_CS/Program.cs
@@ -1,12 +1,12 @@
-﻿using Grpc.Core;
+﻿using System;
+using System.Threading;
+using System.Threading.Tasks;
+using Grpc.Core;
 using Grpc.Net.Client;
 using MegaCorp;
 using ProtoBuf.Grpc;
 using ProtoBuf.Grpc.Client;
 using Shared_CS;
-using System;
-using System.Threading;
-using System.Threading.Tasks;
 
 namespace Client_CS
 {
@@ -21,6 +21,7 @@ namespace Client_CS
             Console.WriteLine(result.Result); // 48
 
             var clock = http.CreateGrpcService<ITimeService>();
+            var counter = http.CreateGrpcService<ICounter>();
             using var cancel = new CancellationTokenSource(TimeSpan.FromMinutes(1));
             var options = new CallOptions(cancellationToken: cancel.Token);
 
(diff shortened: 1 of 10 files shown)


mgravell:
couple of things; technically this is a breaking change *but* I think we can ignore that because `ServiceBinder` isn't actually something that folks would subclass normally; however: what is the *use case* for this change? I would hope to see a test that shows the change being useful for something - and I'm honestly not quite sure what this is right now. How do you propose to make use of this, independently of the other changes? If it isn't useful by itself, I'm not sure of the wisdom of splitting it off.

Euan-McVie:
Hi,
The use case is to allow an override to allow fetching the metadata when a singleton is being used.
I have added an example of the singleton registration.
It includes a fake authentication handler that either treats all incoming calls as authenticated or not based on the `AlwaysAuthenticate` flag. This was to avoid setting up proper authentication handlers etc on the server and client side.

**Correct Behaviour**
- If you set the `AlwaysAuthenticate` flag to `true` (the checked in state):
  - All calls are treated as authenticated.
  - Calls to increment the counter pass the authorization check from the attribute.
- If you set the `AlwaysAuthenticate` flag to `false`
  - All calls are treated as unauthenticated (anonymous)
  - Calls to increment the counter fail the authorization check from the attribute.

**Failing Behaviour**
- If you set the `AlwaysAuthenticate` flag to `false` AND comment out the custom binder `ServiceBinderWithServiceResolutionFromServiceCollection` on line 21.
  - All calls are treated as unauthenticated (anonymous)
  - Calls to increment the counter ignore the authorization attribute and are allowed to run.

The root cause of this is detailed in #121 as registering a singleton into gRPC works, it is just the metadata that fails to pick up attributes from the concrete type as it is only the interface that is registered to with your code.

Ideally we would never register the gRPC service as a singleton, but we have a lot of "legacy" implementations that we are looking to add gRPC to and don't want to have to create a wrapping service that just does some passthrough for all of them.

Because of this I would think it is not worth you coming up with a built in, transparent solution to support what is a bit of a bad pattern and so exposing a way of us injecting the behaviour ourselves is a good workaround.

mgravell:
I'm still a little concerned about the lack of a test that shows the difference in behavior - in particular, my concern here is that without a test, we might accidentally *break* the feature without noticing. But: I'm going to merge as-is, with thanks. It probably still warrants a test, though.

PR #143, opened 2020-11-17
Fix ordering of FileDescriptors in reflection response

### LINKED ISSUE #141: Issues with server reflection for code-first
Hi @mgravell , hi @mholo65 ,

I am experimenting with server reflection.
The sample application I am using is this one:
https://github.com/thinktecture/blazor-webassembly-demo

For discovering I am using Martin's CLI tool like this:

```
C:\Users\christianweyer\>dotnet grpc-cli ls https://localhost:5003
ConfTool.Shared.Contracts.TimeService
ConfTool.Shared.Contracts.ConferencesService
grpc.reflection.v1alpha.ServerReflection
```

That actually looks good.
But when looking at the server, there are some issues:

```
info: Grpc.AspNetCore.Server.ServerCallHandler[14]
      Error reading message.
      System.IO.IOException: The request stream was aborted.
       ---> Microsoft.AspNetCore.Connections.ConnectionAbortedException: The HTTP/2 connection faulted.
       ---> Microsoft.AspNetCore.Connections.ConnectionResetException: An existing connection was forcibly closed by the remote host.
       ---> System.Net.Sockets.SocketException (10054): An existing connection was forcibly closed by the remote host.
         at Microsoft.AspNetCore.Server.Kestrel.Transport.Sockets.Internal.SocketAwaitableEventArgs.<GetResult>g__ThrowSocketException|7_0(SocketError e)
         at Microsoft.AspNetCore.Server.Kestrel.Transport.Sockets.Internal.SocketAwaitableEventArgs.GetResult()
         at Microsoft.AspNetCore.Server.Kestrel.Transport.Sockets.Internal.SocketConnection.ProcessReceives()
         at Microsoft.AspNetCore.Server.Kestrel.Transport.Sockets.Internal.SocketConnection.DoReceive()
         --- End of inner exception stack trace ---
         at System.IO.Pipelines.PipeCompletion.ThrowLatchedException()
         at System.IO.Pipelines.Pipe.GetReadResult(ReadResult& result)
         at System.IO.Pipelines.Pipe.GetReadAsyncResult()
         at Microsoft.AspNetCore.Server.Kestrel.Core.Internal.DuplexPipeStream.ReadAsyncInternal(Memory`1 destination, CancellationToken cancellationToken)
         at System.Net.Security.SslStream.ReadAsyncInternal[TIOAdapter](TIOAdapter adapter, Memory`1 buffer)
         at System.IO.Pipelines.StreamPipeReader.ReadAsync(CancellationToken cancellationToken)
         at Microsoft.AspNetCore.Server.Kestrel.Core.Internal.Http2.Http2Connection.ReadInputAsync()
         at System.IO.Pipelines.PipeCompletion.ThrowLatchedException()
         at System.IO.Pipelines.Pipe.GetReadResult(ReadResult& result)
         at System.IO.Pipelines.Pipe.GetReadAsyncResult()
         at Microsoft.AspNetCore.Server.Kestrel.Core.Internal.Http2.Http2Connection.ProcessRequestsAsync[TContext](IHttpApplication`1 application)
         --- End of inner exception stack trace ---
         --- End of inner exception stack trace ---
         at System.IO.Pipelines.PipeCompletion.ThrowLatchedException()
         at System.IO.Pipelines.Pipe.GetReadResult(ReadResult& result)
         at System.IO.Pipelines.Pipe.GetReadAsyncResult()
         at Microsoft.AspNetCore.Server.Kestrel.Core.Internal.Http2.Http2MessageBody.ReadAsync(CancellationToken cancellationToken)
         at Grpc.AspNetCore.Server.Internal.PipeExtensions.ReadStreamMessageAsync[T](PipeReader input, HttpContextServerCallContext serverCallContext, Func`2 deserializer, CancellationToken cancellationToken)
fail: Grpc.AspNetCore.Server.ServerCallHandler[6]
      Error when executing service method 'ServerReflectionInfo'.
      System.IO.IOException: The request stream was aborted.
       ---> Microsoft.AspNetCore.Connections.ConnectionAbortedException: The HTTP/2 connection faulted.
       ---> Microsoft.AspNetCore.Connections.ConnectionResetException: An existing connection was forcibly closed by the remote host.
       ---> System.Net.Sockets.SocketException (10054): An existing connection was forcibly closed by the remote host.
         at Microsoft.AspNetCore.Server.Kestrel.Transport.Sockets.Internal.SocketAwaitableEventArgs.<GetResult>g__ThrowSocketException|7_0(SocketError e)
         at Microsoft.AspNetCore.Server.Kestrel.Transport.Sockets.Internal.SocketAwaitableEventArgs.GetResult()
         at Microsoft.AspNetCore.Server.Kestrel.Transport.Sockets.Internal.SocketConnection.ProcessReceives()
         at Microsoft.AspNetCore.Server.Kestrel.Transport.Sockets.Internal.SocketConnection.DoReceive()
         --- End of inner exception stack trace ---
         at System.IO.Pipelines.PipeCompletion.ThrowLatchedException()
         at System.IO.Pipelines.Pipe.GetReadResult(ReadResult& result)
         at System.IO.Pipelines.Pipe.GetReadAsyncResult()
         at Microsoft.AspNetCore.Server.Kestrel.Core.Internal.DuplexPipeStream.ReadAsyncInternal(Memory`1 destination, CancellationToken cancellationToken)
         at System.Net.Security.SslStream.ReadAsyncInternal[TIOAdapter](TIOAdapter adapter, Memory`1 buffer)
         at System.IO.Pipelines.StreamPipeReader.ReadAsync(CancellationToken cancellationToken)
         at Microsoft.AspNetCore.Server.Kestrel.Core.Internal.Http2.Http2Connection.ReadInputAsync()
         at System.IO.Pipelines.PipeCompletion.ThrowLatchedException()
         at System.IO.Pipelines.Pipe.GetReadResult(ReadResult& result)
         at System.IO.Pipelines.Pipe.GetReadAsyncResult()
         at Microsoft.AspNetCore.Server.Kestrel.Core.Internal.Http2.Http2Connection.ProcessRequestsAsync[TContext](IHttpApplication`1 application)
         --- End of inner exception stack trace ---
         --- End of inner exception stack trace ---
         at System.IO.Pipelines.PipeCompletion.ThrowLatchedException()
         at System.IO.Pipelines.Pipe.GetReadResult(ReadResult& result)
         at System.IO.Pipelines.Pipe.GetReadAsyncResult()
         at Microsoft.AspNetCore.Server.Kestrel.Core.Internal.Http2.Http2MessageBody.ReadAsync(CancellationToken cancellationToken)
         at Grpc.AspNetCore.Server.Internal.PipeExtensions.ReadStreamMessageAsync[T](PipeReader input, HttpContextServerCallContext serverCallContext, Func`2 deserializer, CancellationToken cancellationToken)
         at Grpc.AspNetCore.Server.Internal.HttpContextStreamReader`1.<MoveNext>g__MoveNextAsync|10_0(ValueTask`1 readStreamTask)
         at ProtoBuf.Grpc.Internal.Reshape.AsAsyncEnumerable[T](IAsyncStreamReader`1 reader, CancellationToken cancellationToken)+MoveNext() in /_/src/protobuf-net.Grpc/Internal/Reshape.cs:line 121
         at ProtoBuf.Grpc.Internal.Reshape.AsAsyncEnumerable[T](IAsyncStreamReader`1 reader, CancellationToken cancellationToken)+System.Threading.Tasks.Sources.IValueTaskSource<System.Boolean>.GetResult()
         at ProtoBuf.Grpc.Reflection.ReflectionService.grpc.reflection.v1alpha.IServerReflection.ServerReflectionInfoAsync(IAsyncEnumerable`1 requests, CallContext context)+MoveNext() in /_/src/protobuf-net.Grpc.Reflection/ReflectionService.cs:line 52
         at ProtoBuf.Grpc.Reflection.ReflectionService.grpc.reflection.v1alpha.IServerReflection.ServerReflectionInfoAsync(IAsyncEnumerable`1 requests, CallContext context)+MoveNext() in /_/src/protobuf-net.Grpc.Reflection/ReflectionService.cs:line 52
         at ProtoBuf.Grpc.Reflection.ReflectionService.grpc.reflection.v1alpha.IServerReflection.ServerReflectionInfoAsync(IAsyncEnumerable`1 requests, CallContext context)+System.Threading.Tasks.Sources.IValueTaskSource<System.Boolean>.GetResult()
         at ProtoBuf.Grpc.Internal.Reshape.WriteTo[T](IAsyncEnumerable`1 reader, IServerStreamWriter`1 writer, CancellationToken cancellationToken) in /_/src/protobuf-net.Grpc/Internal/Reshape.cs:line 134
         at ProtoBuf.Grpc.Internal.Reshape.WriteTo[T](IAsyncEnumerable`1 reader, IServerStreamWriter`1 writer, CancellationToken cancellationToken) in /_/src/protobuf-net.Grpc/Internal/Reshape.cs:line 134
         at Grpc.Shared.Server.DuplexStreamingServerMethodInvoker`3.Invoke(HttpContext httpContext, ServerCallContext serverCallContext, IAsyncStreamReader`1 requestStream, IServerStreamWriter`1 responseStream)
         at Grpc.Shared.Server.DuplexStreamingServerMethodInvoker`3.Invoke(HttpContext httpContext, ServerCallContext serverCallContext, IAsyncStreamReader`1 requestStream, IServerStreamWriter`1 responseStream)
         at Grpc.AspNetCore.Server.Internal.CallHandlers.ServerCallHandlerBase`3.<HandleCallAsync>g__AwaitHandleCall|8_0(HttpContextServerCallContext serverCallContext, Method`2 method, Task handleCall)
```

If I then try to get details for one of the services with the CLI tool, I get this error on the client:

```
C:\Users\christianweyer>dotnet grpc-cli ls https://localhost:5003 ConfTool.Shared.Contracts.ConferencesService

Error: Dependency missing: protobuf-net/bcl.proto
```

Any idea what could be wrong?
If you need more data, please let me know.

Thanks for your great work!

mholo65:
- Fixes #141

### DIFF
--- a/src/protobuf-net.Grpc.Reflection/FileDescriptorSetFactory.cs
+++ b/src/protobuf-net.Grpc.Reflection/FileDescriptorSetFactory.cs
@@ -34,89 +34,21 @@ namespace ProtoBuf.Grpc.Reflection
             var serviceContracts = typeof(IGrpcService).IsAssignableFrom(serviceType)
                 ? new HashSet<Type> { serviceType }
                 : ContractOperation.ExpandInterfaces(serviceType);
+            var schemaGenerator = new SchemaGenerator
+            {
+                BinderConfiguration = binderConfiguration,
+                ProtoSyntax = ProtoSyntax.Proto3
+            };
 
             foreach (var serviceContract in serviceContracts)
             {
-                if (!binderConfiguration.Binder.IsServiceContract(serviceContract, out string? serviceName)) continue;
-
-                var serviceDescriptor = new ServiceDescriptorProto
-                {
-                    Name = ServiceBinder.GetNameParts(serviceName, serviceType, out var package),
-                };
-
-                var dependencies = new HashSet<string>();
-                foreach (var op in ContractOperation.FindOperations(binderConfiguration, serviceContract, null))
-                {
-                    // TODO: Validate op
-                    if (TryGetType(binderConfiguration, op.From, fileDescriptorSet, out var inputType, out var inputFile)
-                        && TryGetType(binderConfiguration, op.To, fileDescriptorSet, out var outputType, out var outputFile))
-                    {
-                        serviceDescriptor.Methods.Add(new MethodDescriptorProto
-                        {
-                            Name = op.Name,
-                            InputType = inputType,
-                            OutputType = outputType,
-                            ClientStreaming = op.MethodType == MethodType.ClientStreaming ||
-                                              op.MethodType == MethodType.DuplexStreaming,
-                            ServerStreaming = op.MethodType == MethodType.ServerStreaming ||
-                                              op.MethodType == MethodType.DuplexStreaming
-                        });
-
-                        dependencies.Add(inputFile);
-                        dependencies.Add(outputFile);
-                    }
-                }
-
-                var fileDescriptor = new FileDescriptorProto
-                {
-                    Name = serviceName + ".proto",
-                    Services = { serviceDescriptor },
-                    Syntax = "proto3",
-                    Package = package
-                };
-
-                foreach (var dependency in dependencies)
-                {
-                    fileDescriptor.Dependencies.Add(dependency);
-                    fileDescriptor.AddImport(dependency, true, default);
-                }
-
-                fileDescriptorSet.Files.Add(fileDescriptor);
-            }
-        }
-
-        private static bool TryGetType(BinderConfiguration binderConfiguration, Type type, FileDescriptorSet fileDescriptorSet, out string fullyQualifiedName, out string descriptorProto)
-        {
-            var typeName = type.Name;
-            var fileName = type.FullName + ".proto";
-            var fileDescriptor = fileDescriptorSet.Files.SingleOrDefault(f => f.Name.Equals(fileName, StringComparison.Ordinal));
-
-            TypeModel model = binderConfiguration.MarshallerCache.TryGetFactory(type) is ProtoBufMarshallerFactory factory ? factory.Model : RuntimeTypeModel.Default;
-
-            if (fileDescriptor is null)
-            {
-                var schema = model.GetSchema(type, ProtoSyntax.Proto3);
-
+                if (!binderConfiguration.Binder.IsServiceContract(serviceContract, out var serviceName)) continue;
+                
+                var schema = schemaGenerator.GetSchema(serviceContract);
                 using var reader = new StringReader(schema);
 
-                fileDescriptorSet.Add(fileName, includeInOutput: true, reader);
-                fileDescriptor = fileDescriptorSet.Files.SingleOrDefault(f => f.Name.Equals(fileName, StringComparison.Ordinal));
-                if (fileDescriptor is null)
-                {
-                    fullyQualifiedName = descriptorProto = "";
-                    return false;
-                }
+                fileDescriptorSet.Add(serviceName + ".proto", includeInOutput: true, reader);
             }
-
-            var msgType = fileDescriptor.MessageTypes.SingleOrDefault(m => m.Name.Equals(typeName, StringComparison.OrdinalIgnoreCase));
-            if (msgType is null)
-            {
-                fullyQualifiedName = descriptorProto = "";
-                return false;
-            }
-            descriptorProto = fileDescriptor.Name;
-            fullyQualifiedName = "." + fileDescriptor.Package + "." + msgType.Name;
-            return true;
         }
     }
 }
--- a/src/protobuf-net.Grpc.Reflection/ReflectionService.cs
+++ b/src/protobuf-net.Grpc.Reflection/ReflectionService.cs
@@ -159,20 +159,34 @@ namespace ProtoBuf.Grpc.Reflection
                 {
                     return 1;
                 }
-                if (left.Dependencies.Contains(right.Name))
+                if (GetTransitiveDependencies(left).Contains(right.Name))
                 {
                     return 1;
                 }
-                if (right.Dependencies.Contains(left.Name))
+                if (GetTransitiveDependencies(right).Contains(left.Name))
                 {
                     return -1;
                 }
-                if (left.Name == right.Name)
-                {
-                    return 0;
-                }
 
-                return -1;
+                return string.Compare(left.Name, right.Name, StringComparison.Ordinal);
+
+                static IReadOnlyCollection<string> GetTransitiveDependencies(FileDescriptorProto? descriptor)
+                {
+                    if (descriptor is null)
+                    {
+                        return Array.Empty<string>();
+                    }
+
+                    var dependencies = new List<string>();
+
+                    foreach (var dependency in descriptor.GetDependencies())
+                    {
+                        dependencies.Add(dependency.Name);
+                        dependencies.AddRange(GetTransitiveDependencies(dependency));
+                    }
+
+                    return dependencies;
+                }
             }
         }
     }
--- a/src/protobuf-net.Grpc.Reflection/ReflectionService.cs
+++ b/src/protobuf-net.Grpc.Reflection/ReflectionService.cs
@@ -155,7 +155,7 @@ namespace ProtoBuf.Grpc.Reflection
                 {
                     return right is null ? 0 : -1;
                 }
-                else if (right is null)
+                if (right is null)
                 {
                     return 1;
                 }
@@ -172,7 +172,7 @@ namespace ProtoBuf.Grpc.Reflection
                     return 0;
                 }
 
-                return 1;
+                return -1;
             }
         }
     }
--- a/tests/protobuf-net.Grpc.Reflection.Test/ReflectionServiceTests.cs
+++ b/tests/protobuf-net.Grpc.Reflection.Test/ReflectionServiceTests.cs
@@ -1,0 +1,145 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Threading.Tasks;
+using Google.Protobuf.Reflection;
+using grpc.reflection.v1alpha;
+using ProtoBuf;
+using ProtoBuf.Grpc.Configuration;
+using ProtoBuf.Grpc.Reflection;
+using Xunit;
+
+namespace protobuf_net.Grpc.Reflection.Test
+{
+    using ReflectionTest;
+
+    public class ReflectionServiceTests
+    {
+        [Theory]
+        [MemberData(nameof(Dependencies))]
+        public async Task ShouldIncludeDependenciesInCorrectOrder(Type service, string symbolName, string[] expectedDescriptors)
+        {
+            IServerReflection reflectionService = new ReflectionService(service);
+
+            await foreach (var response in reflectionService.ServerReflectionInfoAsync(GetRequest()))
+            {
+                var fileDescriptors = response
+                    .FileDescriptorResponse
+                    .FileDescriptorProtoes
+                    .Select(x => Serializer.Deserialize<FileDescriptorProto>(x.AsSpan()))
+                    .ToArray();
+
+                Assert.Equal(expectedDescriptors, fileDescriptors.Select(x => x.Name));
+            }
+
+            async IAsyncEnumerable<ServerReflectionRequest> GetRequest()
+            {
+                yield return new ServerReflectionRequest
+                {
+                    FileContainingSymbol = symbolName
+                };
+
+                await Task.CompletedTask;
+            }
+        }
+
+        public static IEnumerable<object[]> Dependencies => new[]
+        {
+            new object[]
+            {
+                typeof(BclService),
+                ".ReflectionTest.BclService",
+                new[]
+                {
+                    "ProtoBuf.Grpc.Internal.Empty.proto", // TODO: Maybe Google well-known should have correct name?
+                    "protobuf-net/bcl.proto",
+                    "ReflectionTest.BclMessage.proto",
+                    "ReflectionTest.BclService.proto"
+                }
+            },
+            new object[]
+            {
+                typeof(Nested),
+                ".ReflectionTest.Nested",
+                new[]
+                {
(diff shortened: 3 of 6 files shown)


tests/protobuf-net.Grpc.Reflection.Test/ReflectionServiceTests.cs
```diff
@@ -0,0 +1,145 @@
...
+
+        public static IEnumerable<object[]> Dependencies => new[]
+        {
+            new object[]
+            {
+                typeof(BclService),
+                ".ReflectionTest.BclService",
+                new[]
+                {
+                    "ProtoBuf.Grpc.Internal.Empty.proto", // TODO: Maybe Google well-known should have correct name?
```
mholo65, line 54: `Empty` doesn't seem to use Google well-known.. Should it?

tests/protobuf-net.Grpc.Reflection.Test/ReflectionServiceTests.cs
```diff
@@ -0,0 +1,145 @@
...
+                }
+            },
+            new object[]
+            {
+                typeof(Nested),
+                ".ReflectionTest.Nested",
+                new[]
+                {
+                    "ReflectionTest.Three.proto", // This includes Two and One
+                    "ReflectionTest.Four.proto", // This includes Three, Two and One
```
mholo65, line 67: These can become messy when file descriptors contains multiple messages... Especially if one would like to use the reflection response for e.g. generating `.proto` files.

src/protobuf-net.Grpc.Reflection/ReflectionService.cs
```diff
@@ -172,7 +172,7 @@ public int Compare(FileDescriptorProto? left, FileDescriptorProto? right)
                     return 0;
                 }
 
-                return 1;
+                return -1;
```
mholo65, line 175: This was the bug :)

tests/protobuf-net.Grpc.Reflection.Test/ReflectionServiceTests.cs
mholo65: Btw, @mgravell, any reason we shouldn't switch to use `SchemaGenerator` instead in the `FileDescriptorSetFactory`? This would include everything in one FileDescriptor per service..

src/protobuf-net.Grpc.Reflection/ReflectionService.cs
mgravell: I don't think this is a valid fix; consider two arbitrary descriptors without a dependency relation and with different names. With this as *either* 1 or -1, we get the scenario where `Compare(x,y)` would return -1 **and** `Compare(y,x)` would return -1, which is *not* a valid comparer.

IMO the lines 170-175 should be simply `return string.Compare(left.Name, right.Name)` ?

tests/protobuf-net.Grpc.Reflection.Test/ReflectionServiceTests.cs
mgravell: I've eyeballed `FileDescriptorSetFactory` and I'm not entirely sure what you mean here; happy to consider changes, but: it isn't leaping out at me. Possibly I just need more coffee.

tests/protobuf-net.Grpc.Reflection.Test/ReflectionServiceTests.cs
mgravell: yeah, that's just wrong; we should generate the well-known name

src/protobuf-net.Grpc.Reflection/ReflectionService.cs
mgravell:
```suggestion
                return string.Compare(left.Name, right.Name);
```

mholo65: String comparison is not needed in this case since this comparer is used to sort FileDescriptors in correct order. In gRPC reflection it is expected that FileDescriptors are received in correct order. Any dependencies must preceed a descriptor that has a depenency on it.

This fix is *should* work for transitive dependencies, as it will put Left on the "left" side of Right, thus Left will then be compared with the dependencies of Right. One could also consider a recursive dependency check in lines 162 and 166.

mgravell: in that case, isn't `return 0;` the more appropriate fix, which any stable sort implementation will interpret as "leave it in the original order"? note that LINQ `OrderBy` guarantees stable sort, but IIRC `List<T>` and `T[]` in-place sort *do not*

mholo65: Yes a `return 0;` and do a recursive dependency lookup here and here:
https://github.com/protobuf-net/protobuf-net.Grpc/blob/3ebd1a07c4520fdcc0fbf6271e12dbfc3f5160eb/src/protobuf-net.Grpc.Reflection/ReflectionService.cs#L162

https://github.com/protobuf-net/protobuf-net.Grpc/blob/3ebd1a07c4520fdcc0fbf6271e12dbfc3f5160eb/src/protobuf-net.Grpc.Reflection/ReflectionService.cs#L166

Should be a correct fix.

tests/protobuf-net.Grpc.Reflection.Test/ReflectionServiceTests.cs
mholo65: They both do similar things, although `SchemaGenerator` only generates the `.proto` schema, Reflection service should return `FileDescriptors`. But we could use `SchemaGenerator` as base, and then let `protobuf-net.Reflection` convert it to a `FileDescriptor`. This way we would get one "self-contained" `FileDescriptor` per service (not really, since well-known types will be available as separate descriptors).

src/protobuf-net.Grpc.Reflection/ReflectionService.cs
mholo65: Fixed.

tests/protobuf-net.Grpc.Reflection.Test/ReflectionServiceTests.cs
mholo65: Fixed by using SchemaGenerator

tests/protobuf-net.Grpc.Reflection.Test/ReflectionServiceTests.cs
mholo65: I went the SchemaGenerator route

src/protobuf-net.Grpc.Reflection/FileDescriptorSetFactory.cs
```diff
@@ -34,89 +34,21 @@ private static void Populate(FileDescriptorSet fileDescriptorSet, Type serviceTy
...
+            };
 
-            TypeModel model = binderConfiguration.MarshallerCache.TryGetFactory(type) is ProtoBufMarshallerFactory factory ? factory.Model : RuntimeTypeModel.Default;
-
-            if (fileDescriptor is null)
+            foreach (var serviceContract in serviceContracts)
             {
-                var schema = model.GetSchema(type, ProtoSyntax.Proto3);
-
+                if (!binderConfiguration.Binder.IsServiceContract(serviceContract, out var serviceName)) continue;
```
mholo65, line 45: Maybe we could add service name as out param to `SchemaGenerator.GetSchema()`? Then we wouldn't need to call this here. It's also called in schema generator.

tests/protobuf-net.Grpc.Reflection.Test/ReflectionServiceTests.cs
```diff
@@ -63,8 +63,8 @@ async IAsyncEnumerable<ServerReflectionRequest> GetRequest()
                 ".ReflectionTest.Nested",
                 new[]
                 {
-                    "ReflectionTest.Three.proto", // This includes Two and One
-                    "ReflectionTest.Four.proto", // This includes Three, Two and One
+                    "ReflectionTest.Four.proto", // This includes Two and One
+                    "ReflectionTest.Three.proto", // This includes Three, Two and One
```
mgravell: Three includes three?

mgravell (approved):
Transitive compare looks good; was the other change intended?

tests/protobuf-net.Grpc.Reflection.Test/ReflectionServiceTests.cs
mholo65: This is an outdated diff. This only returns one `.proto` file, `ReflectionTest.Nested.proto`.

mholo65:
> 
> 
> Transitive compare looks good; was the other change intended?

Yep, all changes was intentional. Using `SchemaGenerator` cleaned up the code somewhat, as `FileDescriptorSetFactory` was doing almost same stuff.

mholo65:
It now occurred to me that this will most probably NOT work if one of the message dependencies (transitive or not) is located in a different package (namespace). Isn't it so that SchemaGenerator will generate one file per package?

mgravell:
If they're in different packages, they need to be in different schema files
- if that is the question. So they'd have to come from an import. I have
recently started adding a mechanism in v3 to allow us to declare external
files, so that we can generate the right imports - any use? But:
restricting to a single package is indeed necessary.

mholo65:
> I have recently started adding a mechanism in v3 to allow us to declare external files, so that we can generate the right imports - any use?

Ah, so there is (or will be) an API for generating schema for imports also. Haven't tried, but I guess we don't get schema for imports now, because SchemaGenerator only returns a string.

Anyhow, I will make sure FileDescriptors are also included for imports from other packages before we can merge this PR.

mholo65:
Updated tests to verify this, and it seems that `TypeModel` doesn't care about the namespace of the message types. It will happily add all dependent message types in the same schema generated for the service. What dictates the actual `package` is what's passed in the `SchemaGenerationOptions`.

I think this is ok now.

mgravell:
merged with thanks

PR #157, opened 2021-01-31
fix premature cancellation in CallContext.FullDuplexAsync

i3arnon:


mgravell:
The input is appreciated. I'm going to have to check carefully - unless I'm mistaken this is different to the code we discussed earlier, and I *think* is the working version. Or maybe I am being blind on a Sunday evening. I will look in depth on Monday, time permitting.

mgravell:
Sorry for delay on this - I had a bout of coronavirus, which wiped me out for a while. I don't *think* anything is wrong here; the `SwapOut` call is intended to ensure that `allDone` is only disposed in the scenarios in which control is handed to `BothAsync`. Is there a demonstrable scenario in which you're seeing something unexpected happen?

i3arnon:
> Sorry for delay on this - I had a bout of coronavirus, which wiped me out for a while

No worries. I hope you feel better..

> the `SwapOut` call is intended to ensure that `allDone` is only disposed in the scenarios in which control is handed to `BothAsync`

Indeed, that's right.
My eyes are accustomed for hunting `using`/`finally` blocks without `await`, but less for `ref` arguments.

PR #165, opened 2021-03-09
Prevent unawaited task (which could fault) from being dropped

mgravell:
If the call to `producer(...)` throws, we may end up dropping a Task (the consumer) on the floor, which could itself fault (probably due to cancellation)

1. make sure that we observe the task in this scenario
2. cancel the composite CTS *before* we try this, so we've signalled everything


deanward81 (approved):
Looks sane, will give it a spin in my deployment branch 👍

PR #188, opened 2021-06-25
Fix an unobserved task exception

deanward81:
Addresses an unobserved exception that occurs when an `ObjectDisposedException` is raised after a `CancellationToken` is cancelled by only cancelling if the token has not been previously cancelled. The default `catch` will fire for an `OperationCanceledException` and then attempt to cancel the same `CancellationToken` that was just cancelled.

### DIFF
--- a/src/protobuf-net.Grpc/Internal/Reshape.cs
+++ b/src/protobuf-net.Grpc/Internal/Reshape.cs
@@ -520,7 +520,10 @@ namespace ProtoBuf.Grpc.Internal
             }
             catch
             {
-                allDone.Cancel();
+                if (!allDone.IsCancellationRequested)
+                {
+                    allDone.Cancel();
+                }
                 throw;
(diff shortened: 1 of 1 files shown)


src/protobuf-net.Grpc/Internal/Reshape.cs
```diff
@@ -520,7 +520,10 @@ await foreach (var value in request.WithCancellation(allDone.Token).ConfigureAwa
             }
             catch
             {
-                allDone.Cancel();
+                if (!allDone.IsCancellationRequested)
```
mgravell, line 523: the fact that we need to test this suggests that there's competition, which suggests that this is itself an inherent race-condition; as such, I suggest putting a `try`/`catch` around the actual `Cancel()`; thoughts?

mgravell (approved):

PR #199, opened 2021-10-14
Fix a couple of semi-edge-cases

lanfeust69:
These could result in not binding a service at all, when only one method is at fault, or throwing unexpectedly on the client side.


mgravell (approved):

mgravell:
this was long overdue - entirely my fault; merged with thanks - should be in 1.1

PR #200, opened 2021-10-14
Allow binding all operations to top-level service

lanfeust69:
As explained in the inline documentation :

This is useful if several services implement a common interface, for example :
```
    interface IMonitoredService { ServiceReport GetStatus(); }
```
In this case, we don't want to call *any* service implementing this, but on the contrary to have *all* services be queriable individually on `/MySpecificService/GetStatus`.

### DIFF
--- a/src/protobuf-net.Grpc/Configuration/BinderConfiguration.cs
+++ b/src/protobuf-net.Grpc/Configuration/BinderConfiguration.cs
@@ -1,4 +1,4 @@
-﻿using Grpc.Core;
+using Grpc.Core;
 using ProtoBuf.Grpc.Internal;
 using System;
 using System.Collections.Generic;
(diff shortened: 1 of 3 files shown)


mgravell:
I'm still unclear on the use-case of this one; I *think* this might be the same underlying scenario as https://github.com/protobuf-net/protobuf-net/issues/859, in which case: I think we can address it from the approach discussed there. I'm going to try putting a PR together for that one.

mgravell:
I suspect that what this scenario is describing is now better handled via `[SubService]`:

``` c#
[SubService]
interface IMonitoredService { ServiceReport GetStatus(); }
[Service]
interface IMySpecificService : IMonitoredService { ... }
[Service]
interface IMyOtherService : IMonitoredService { ... }
```

Closing this on that basis. If I've misunderstood: let me know - we can always reopen.

lanfeust69:
@mgravell : Yes it looks like the new feature addresses the issue. However, I won't be able to confirm for sure, as I'm no longer involved with the project that needed this.

PR #209, opened 2021-12-05
Reflection - allow generating multi services with their shared classes defined once.

meirkr:
allow generating multi services with their shared classes defined once.
Suppose there is a service which gets some params:

Suppose we have following two services:
```
    [Service]
    interface IVehicleRentService
    {
        Task<VehicleRentResponse> RentVehicle(VehicleRentRequest vehicleRentRequest);
    }

    [Service]
    interface IToxedoRentService
    {
        Task<ToxedoRentResponse> RentToxedo(ToxedoRentRequest toxedoRentRequest);
    }

    [ProtoContract]
    public class ToxedoRentRequest
    {
        [ProtoMember(1)]
        public RenterPersonalInfo RenterPersonalInfo { get; set; }
        [ProtoMember(2)]
        public RenterFavorites RenterFavorites { get; set; }
        [ProtoMember(3)]
        public Purpose RentPurpose { get; set; }

        public enum Purpose
        {
            Business,
            Joy,
            Travel,
            // etc
        };
    }
    [ProtoContract]
    public class VehicleRentRequest
    {
        [ProtoMember(1)]
        public RenterPersonalInfo RenterPersonalInfo { get; set; }
        [ProtoMember(2)]
        public RenterFavorites RenterFavorites { get; set; }
        [ProtoMember(3)]
        public VehicleLicense[]  LicenseTypes { get; set; }
        
        public enum VehicleLicense
        {
            Private,
            Motorcycle,
            Track,
            Bus,
            Semitrailer,
        };        
    }    

    [ProtoContract]
    public class RenterFavorites
    {
        public enum PreferredColor
        {
            DontCare,
            Red,
            White,
            // etc
        };
        public enum PreferredStyle
        {
            DontCare,
            Cool,
            Solid,
            // etc
        };

        [ProtoMember(1)]
        public PreferredColor RenterPreferredColor { get; set; }
        [ProtoMember(2)]
        public PreferredStyle RenterPreferredStyle { get; set; }
    }

    internal class RenterPersonalInfo
    {
        [ProtoMember(1)]
        public string Id { get; set; }
        [ProtoMember(2)]
        public string FirstName { get; set; }
        [ProtoMember(3)]
        public string LastName { get; set; }
        [ProtoMember(4)]
        public uint Age { get; set; }
    }
```

Since there are several classes which are resued between the services, it won't be good generating different proto files.
The outcome will be that their code gen will produce duplicated classes, which cannot be reused. 
A client's  developers will need to fill in the inputs of the two services separately, while nothing can be reused.
Hard work and a risk for mistakes.
Better will be if both services schema(s) will finally produce code which can be reused.

This PR lets that happen by producing a single proto file representing several services (currently only same namespace/package is supported) and all the shared classes are defined only once.
That will let the developers **filling the shared information only once** and providing that same data as (reused) input to the different services to be handled separately. 

Added more unit tests to cover this feature.
Added an addition unit test of exception in case schema gets a non service contract interface/class.

### DIFF
--- a/src/protobuf-net.Grpc.Reflection/SchemaGenerator.cs
+++ b/src/protobuf-net.Grpc.Reflection/SchemaGenerator.cs
@@ -26,14 +26,26 @@ namespace ProtoBuf.Grpc.Reflection
         /// <summary>
         /// Get the .proto schema associated with a service contract
         /// </summary>
+        /// <typeparam name="TService">The service type to generate schema for.</typeparam>
         /// <remarks>This API is considered experimental and may change slightly</remarks>
         public string GetSchema<TService>()
             => GetSchema(typeof(TService));
 
         /// <summary>
+        /// Get the .proto schema associated with a service contract
+        /// </summary>
+        /// <param name="contractType">The service type to generate schema for.</param>
+        /// <remarks>This API is considered experimental and may change slightly</remarks>
+        public string GetSchema(Type contractType)
+            => GetSchema(new [] {contractType});
+
+        /// <summary>
         /// Get the .proto schema associated with multiple service contracts
         /// </summary>
-        /// <remarks>This API is considered experimental and may change slightly</remarks>
+        /// <param name="contractTypes">Array (or params syntax) of service types to generate schema for.</param>
+        /// <remarks>This API is considered experimental and may change slightly
+        /// All types will be generated into single schema.
+        /// All the shared classes the services use will be generated only once for all of them.</remarks>
         public string GetSchema(params Type[] contractTypes)
         {
             string globalPackage = "";
--- a/src/protobuf-net.Grpc.Reflection/SchemaGenerator.cs
+++ b/src/protobuf-net.Grpc.Reflection/SchemaGenerator.cs
@@ -35,7 +35,10 @@ namespace ProtoBuf.Grpc.Reflection
         /// Get the .proto schema associated with a service contract
         /// </summary>
         /// <param name="contractType">The service type to generate schema for.</param>
-        /// <remarks>This API is considered experimental and may change slightly</remarks>
+        /// <remarks>This API is considered experimental and may change slightly.
+        /// ATTENTION! although the 'GetSchema(params Type[] contractTypes)' covers also a case of 'GetSchema(Type contractType)',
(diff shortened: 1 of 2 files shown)


mgravell:
Any chance of a small example in the description (i.e. here, not the source), that shows what the intent is here?

meirkr:
edited with an example.

mgravell:
Thanks, much better context!

meirkr:
Hi
Did you have time to review?
Is there something I can do to advance the delivery process?

mgravell:
I'd rather avoid the API break since it doesn't seem *necessary*; instead of dropping the old method, could we instead have the old method call the new method, passing a single `Type`?

``` c#
public string GetSchema(Type contractType) => GetSchema(new[] {contractType});
public string GetSchema(params Type[] contractTypes) ...
```

meirkr:
Actually, this is what I did initially.
Later I have realized that in compile time it is backward compatible, as "params" can get a single "Type" same as the existing method.
In case there is a chance that someone will upgrade to a new version without recompile, there is indeed a compatibility issue loading that new method in runtime when code was compiled with previous version.
Do you think there could be such case that someone upgrades somehow without compiling their code with this library? I guess they all get the libraries using Nuget, so they actually recompile with the new Nuget package anyway. don't they?
Again - their code should not be modified accordingly.

mgravell:
> Do you think there could be such case that someone upgrades somehow without compiling their code with this library?

Yes, absolutely 100% this happens.

In particular, what tends to happen is:

- some other library, let's call it Foo, refs us, and consumes an API (Foo could be a public nuget package, for example ServiceStack, or one of the Microsoft bits, or it could be an internal private company comms lib)
- some application takes a ref on us, and Foo
- at some point, they do a library update and get a new version of us, perhaps because of *another* library they're refing
- and now Foo explodes at runtime with a Missing method exception, because nobody rebuilt Foo, and it is looking for a method that no longer exists

That's why we only do binary compatibility breaks in "majors" - at least people have a clue that things could break - and only for good reasons. In this case, it is absolutely avoidable, so: we should avoid it.

meirkr:
Good to know about such scenarios.
I'll push a change accordingly.

Issue #202, opened 2021-10-28
Parallel test execution leads to non-concurrent collection problems


molson194:
We have created one code-first GRPC service, but when we try to run multiple integration tests at the same time, we hit the following error for half of our tests.

    fail ProtoBuf.Grpc.Server.ServicesExtensions.CodeFirstServiceMethodProvider[0]
    Operations that change non-concurrent collections must have exclusive access.
    A concurrent update was performed on this collection and corrupted its state.
    The collection's state is no longer correct.

The other half of our tests have this message (showing that they succeed to set up the ProbeService)

     info: ProtoBuf.Grpc.Server.ServicesExtensions.CodeFirstServiceMethodProvider[0] RPC services being provided by ProbeService: 1

Our parallel test execution happens at the method level

    [assembly: Parallelize(Scope = ExecutionScope.MethodLevel, Workers = 0)]

During test setup, we instantiate an in-memory ASP.NET Core app using a [custom web application factory](https://docs.microsoft.com/en-us/aspnet/core/test/integration-tests?view=aspnetcore-5.0#customize-webapplicationfactory)

    using var customWebFactory = new CustomWebAppFactory();
    var client = customWebFactory.Channel.CreateGrpcService<IProbeService>(); // This is where the failure happens

Inside our web application factory, we expose the GRPC channel as a property.

    Channel = GrpcChannel.ForAddress($"http://localhost:{grpcPort}/", new GrpcChannelOptions
    {
        HttpClient = CreateClient(),
    });

Our probe service interface looks like this.

    [ServiceContract]
    public interface IProbeService
    {
        public Task<ProbeResponse> Probe(ProbeRequest request);
    }

And our implementation of IProbeService looks like this

    public class ProbeService : IProbeService
    {
        private readonly ILogger<ProbeService> _logger;

        public ProbeService(ILogger<ProbeService> logger)
        {
            _logger = logger;
        }

        public Task<ProbeResponse> Probe(ProbeRequest request)
        {
            _logger.Probe(request.Value);
            return Task.FromResult(new ProbeResponse(request.Value));
        }
    }

It looks like the CodeFirstServiceMethodProvider cannot handle the parallel test execution. For now, I've worked around the issue by running these tests sequentially, but eventually I'd love to run them in parallel. Any thoughts on how to fix the concurrent access issues?

molson194:
[...]

PR #215, opened 2022-01-12
Remove default static properties for thread safety

molson194:


mgravell:
In the state presented, that is massively breaking, so: I'd need very compelling reasons to consider that - i.e. the API is fundamentally untenable.

So: let's take a step back. What are we actually seeing here? Many of these types.are internally immutable. The others *should* be thread-safe; if we've failed somewhere with that, my strong preference here is "let's fix that". I'm going to re-read the original issue #202, but it sounds like the main problem here is `ServiceBinder._map`. I'll take some time this morning to have a look at what this needs, but: I am strongly against blindly removing these properties. The design is intentional; if there's a bug: we fix the bug.

mgravell:
Closing as unmergeable (massively breaking); ref: #216 instead for alternative fix

PR #216, opened 2022-01-12
ServiceBinder may end up being used concurrently; make sure the internal data supports this

### LINKED ISSUE #202: Parallel test execution leads to non-concurrent collection problems

mgravell:
fix #202

also makes the interface-map `static`; this is a view on unchanging IL metadata without interpretation - it isn't instance-specific

### DIFF
--- a/src/protobuf-net.Grpc/Configuration/ServiceBinder.cs
+++ b/src/protobuf-net.Grpc/Configuration/ServiceBinder.cs
@@ -1,7 +1,8 @@
-﻿using System;
+﻿using ProtoBuf.Grpc.Internal;
+using System;
+using System.Collections.Concurrent;
 using System.Collections.Generic;
 using System.Reflection;
-using ProtoBuf.Grpc.Internal;
 
 namespace ProtoBuf.Grpc.Configuration
 {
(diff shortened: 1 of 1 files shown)


molson194:
Thanks for making this change 👍

PR #217, opened 2022-01-27
Extra space removed after the url

zgabi:
It worked earlier when I compiled version 1.0.136 in VS 2019, but now it causes a lot of build errors in VS2022... even that state, too. So probably something was fixed in VS which is not trimming the url anymore.


mgravell:
Oops; thanks

PR #232, opened 2022-03-31
Grpclite initial stab

mgravell:


ChristianWeyer:
HE'S BACK!!! 🥳

ChristianWeyer:
I don't need to tell anyone... I can tell you my wanted features, anyway ;-)

mgravell:
If you have relevant thoughts: sure, chime in

ChristianWeyer:
I still think that code-first JSON/REST transcoding would be a killer. https://github.com/grpc/grpc-dotnet/issues/167

mgravell:
well, this has no pretension to be HTTP, so: not REST; JSON is easy enough - the marshaller has always been pluggable - I just doesn't see the crossover with REST; have more context here?

ChristianWeyer:
Ah, it is not related to this PR, sorry.
The JSON transcoding will land in .NET 7, with contract-first. And it would be a great addition to this lib here for the code-first approach.
Sorry for the confiusion.

mgravell:
Gotcha. I'll need to catch up on what is happening there. In theory, JSON
transcode should work fine with code-first, as long as the JSON matches
your types, potentially annotated for whatever JSON tool they're using. Or
are they looking at the inbuilt Google JSON accent in protobuf?

ChristianWeyer:
I guess @JamesNK could chime in here.

JamesNK:
I doubt gRPC JSON trancoding will work with protobuf-net. It makes heavy use of Google.Protobuf.Reflection. The whole feature is based around [annotating the proto file](https://cloud.google.com/endpoints/docs/grpc-service-config/reference/rpc/google.api#google.api.HttpRule) which doesn't apply to protobuf-net.

There might be a chance that you could have an adapter that exposes Google.Protobuf.Reflection types, or implements its interfaces to enable it to work, but I have looked at it closely.

> as long as the JSON matches your types

Transcoding doesn't do standard POCO serialization. It follows the mapping at https://developers.google.com/protocol-buffers/docs/proto3#json (although I've made int64 serialize as longs by default instead of strings)

mgravell:
@JamesNK interesting, thanks; how open would you be to a potential injection point, around the time of "hey, we've got JSON, let's see what we can do with that"? If you're even remotely open to that kind of thing, I'll happily do the work to see if we can support offloading the JSON piece, but if that is a non-starter: it really isn't worth me looking :)

As a side note: gRPC *isn't even tied to protobuf* - although it is the default; so having the transcode hard tied to the protobuf model sounds like an avoidable thing to me. I entirely agree it makes sense for the *default* implementation to do that, though.

I guess it also depends which route you're using here, by which I mean: "JSON -> object-model" Vs "JSON -> protobuf -> (marshaller) -> object-model"

PR #236, opened 2022-04-13
Support IObservable input/output

mgravell:
/cc @Scottkane - this might even work! (should be on myget shortly)

### DIFF
--- a/src/protobuf-net.Grpc/Configuration/ServerBinder.cs
+++ b/src/protobuf-net.Grpc/Configuration/ServerBinder.cs
@@ -52,7 +52,7 @@ namespace ProtoBuf.Grpc.Configuration
                 var bindCtx = new ServiceBindContext(serviceContract, serviceType, state, binderConfiguration.Binder);
                 foreach (var op in ContractOperation.FindOperations(binderConfiguration, serviceContract, this))
                 {
-                    if (ServerInvokerLookup.TryGetValue(op.MethodType, op.Context, op.Result, op.Void, out var invoker)
+                    if (ServerInvokerLookup.TryGetValue(op.MethodType, op.Context, op.Arg, op.Result, op.Void, out var invoker)
                         && AddMethod(op.From, op.To, op.Name, op.Method, op.MethodType, invoker, bindCtx,
                         serviceContractSimplifiedExceptions || op.Method.IsDefined(typeof(SimpleRpcExceptionsAttribute))
                         ))
--- a/src/protobuf-net.Grpc/Internal/ContractOperation.cs
+++ b/src/protobuf-net.Grpc/Internal/ContractOperation.cs
@@ -17,6 +17,7 @@ namespace ProtoBuf.Grpc.Internal
         public MethodInfo Method { get; }
         public MethodType MethodType { get; }
         public ContextKind Context { get; }
+        public ResultKind Arg { get; }
         public ResultKind Result { get; }
         public VoidKind Void { get; }
         public bool VoidRequest => (Void & VoidKind.Request) != 0;
@@ -25,7 +26,7 @@ namespace ProtoBuf.Grpc.Internal
         public override string ToString() => $"{Name}: {From.Name}=>{To.Name}, {MethodType}, {Result}, {Context}, {Void}";
 
         public ContractOperation(string name, Type from, Type to, MethodInfo method,
-            MethodType methodType, ContextKind contextKind, ResultKind resultKind, VoidKind @void)
+            MethodType methodType, ContextKind contextKind, ResultKind arg, ResultKind resultKind, VoidKind @void)
         {
             Name = name;
             From = from;
@@ -33,6 +34,7 @@ namespace ProtoBuf.Grpc.Internal
             Method = method;
             MethodType = methodType;
             Context = contextKind;
+            Arg = arg;
             Result = resultKind;
             Void = @void;
         }
@@ -48,6 +50,7 @@ namespace ProtoBuf.Grpc.Internal
             IAsyncEnumerable,
             IAsyncStreamReader,
             IServerStreamWriter,
+            IObservable,
             CallOptions,
             ServerCallContext,
             CallContext,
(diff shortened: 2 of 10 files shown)


ScottKane:
Awesome, I will give it a test 👍

mgravell:
feed url: https://www.myget.org/F/protobuf-net/api/v3/index.json; build is 1.0.171-gd5a3b43f3d; however, I'm not convinced that they made it to myget - might need manual build

mgravell:
let's see if this gives it more impetus...

ScottKane:
https://github.com/ScottKane/GrpcLiteTest for reference. Note I may have found a bug with kestrel `AllowedHosts`, it seems after first build kestrel ignores `services.Configure<HostFilteringOptions>(options => options.AllowedHosts = new[] { "*" });`. Then you run again and it works. Doesn't seem to happen with `AllowedHosts` being set in appsettings.json

ScottKane:
It still doesn't work inside Unity though, is this being converted to an IAsyncEnumerable under the covers?

ScottKane:
```
Unhandled exception. System.TimeoutException: The operation has timed out.
   at ProtoBuf.Grpc.Lite.Internal.LiteStream`2.ThrowCancelled() in /_/src/protobuf-net.GrpcLite/Internal/LiteStream.cs:line 287
   at ProtoBuf.Grpc.Lite.Internal.LiteStream`2.ProtoBuf.Grpc.Lite.Internal.IStream.Cancel() in /_/src/protobuf-net.GrpcLite/Internal/LiteStream.cs:line 260
--- End of stack trace from previous location ---
   at ProtoBuf.Grpc.Lite.Internal.LiteStream`2.System.Threading.Tasks.Sources.IValueTaskSource<System.Boolean>.GetResult(Int16 token) in /_/src/protobuf-net.GrpcLite/Internal/LiteStream.cs:line 923
   at ProtoBuf.Grpc.Lite.Internal.LiteStream`2.MoveNextPayloadAsync() in /_/src/protobuf-net.GrpcLite/Internal/LiteStream.cs:line 401   at ProtoBuf.Grpc.Lite.Internal.LiteStream`2.<Grpc.Core.IAsyncStreamReader<TReceive>.MoveNext>g__Awaited|8_0(LiteStream`2 stream, ValueTask pending, CancellationTokenRegistration ctr) in /_/src/protobuf-net.GrpcLite/Internal/LiteStream.cs:line 103
   at ProtoBuf.Grpc.Internal.Reshape.AsAsyncEnumerable[T](IAsyncStreamReader`1 reader, CancellationToken cancellationToken)+MoveNext() in /_/src/protobuf-net.Grpc/Internal/Reshape.cs:line 121
   at ProtoBuf.Grpc.Internal.Reshape.AsAsyncEnumerable[T](IAsyncStreamReader`1 reader, CancellationToken cancellationToken)+System.Threading.Tasks.Sources.IValueTaskSource<System.Boolean>.GetResult()
   at System.Linq.AsyncEnumerable.ToObservableObservable`1.<>c__DisplayClass2_0.<<Subscribe>g__Core|0>d.MoveNext() in /_/Ix.NET/Source/System.Linq.Async/System/Linq/Operators/ToObservable.cs:line 50
--- End of stack trace from previous location ---
   at System.Reactive.PlatformServices.ExceptionServicesImpl.Rethrow(Exception exception) in /_/Rx.NET/Source/src/System.Reactive/Internal/ExceptionServicesImpl.cs:line 19
   at System.Reactive.ExceptionHelpers.Throw(Exception exception) in /_/Rx.NET/Source/src/System.Reactive/Internal/ExceptionServices.cs:line 16
   at System.Reactive.Stubs.<>c.<.cctor>b__2_1(Exception ex) in /_/Rx.NET/Source/src/System.Reactive/Internal/Stubs.cs:line 16
   at System.Reactive.AnonymousObserver`1.OnErrorCore(Exception error) in /_/Rx.NET/Source/src/System.Reactive/AnonymousObserver.cs:line 73
   at System.Reactive.ObserverBase`1.OnError(Exception error) in /_/Rx.NET/Source/src/System.Reactive/ObserverBase.cs:line 59
   at System.Linq.AsyncEnumerable.ToObservableObservable`1.<>c__DisplayClass2_0.<<Subscribe>g__Core|0>d.MoveNext()
--- End of stack trace from previous location ---
   at System.Linq.AsyncEnumerable.ToObservableObservable`1.<>c__DisplayClass2_0.<<Subscribe>g__Core|0>d.MoveNext() in /_/Ix.NET/Source/System.Linq.Async/System/Linq/Operators/ToObservable.cs:line 74
--- End of stack trace from previous location ---
   at System.Threading.Tasks.Task.<>c.<ThrowAsync>b__128_1(Object state)
   at System.Threading.QueueUserWorkItemCallbackDefaultContext.Execute()
   at System.Threading.ThreadPoolWorkQueue.Dispatch()
   at System.Threading.PortableThreadPool.WorkerThread.WorkerThreadStart()
   at System.Threading.Thread.StartCallback()
```

It seems as if the pipe is being closed after the initial request, should I be doing something different to make this stream long lived?

mgravell:
No, I don't think there's a IAsyncEnumerable hidden underneath, at least as
far as I know without spelunking 7 layers into some unrelated internals.
That's kinda why I dislike unity things - nothing works quite like you
expect :/

I'll try to take another look tomorrow, but I can't promise.

Re the connection: again, it should "just work". Without seeing it fail, I
could only throw darts in the general direction of reasons.

ScottKane:
Yeah same haha, I wish they had just kept the runtime up to date, looks like I will have to wait until they move to CoreCLR to do anything cool

mgravell:
I checked, and yes: it does look like I'm making use of `IAsyncEnumerable` in a few places; I'll see how hard it would be to revoke that, although it does kinda mean manually writing an async iterator state machine; not impossible, of course

mgravell:
I had a look at how much work would be involved in removing that... it ... isn't good. In theory, it can be done - but rewriting the core deframer to work this way: is a huge headache; I doubt that it is realistic, unless we're absolutely sure that it would be useful (and work), and honestly: I'm not

mgravell:
OK; I'm a glutton for punishment; a new build is en-route to myget which *completely* removes `IAsyncEnumerbale<T>` (it no longer refs `Microsoft.Bcl.AsyncInterfaces`). If this still doesn't work: I don't think it ever will. But: myget: 0.1.7

ScottKane:
Not all hero's wear capes. I will have a test when I get back to my PC. Thank you for all the effort you've put into this for me.

ScottKane:
It seems as if there is still some IAsyncEnumerable stuff going on as the exception still mentions `ProtoBuf.Grpc.Internal.Reshape.AsAsyncEnumerable[T]`. I think to get this working in Unity I'm just going to have to wait for them to move to CoreCLR. Everything works flawlessly outside of Unity though so thank you.

mgravell:
That's weird; it genuinely shouldn't be - unless unity is objecting *purely because the method exists*. Without a runnable example, it is hard for me to say much, though

ScottKane:
I was being stupid, the problem was that the server wasn't being rebuilt properly (nuget isn't supported in Unity so I'm using a third part tool which for some reason hadn't copied over the new version) and it appears to no longer throw that exception.

I still seem to only get the first message and then nothing more which is strange, trying to put together a smaller sample but for some reason with the exact same code I'm getting connection refused, even though it works in another project.

ScottKane:
Ok so here's a sample you can look at if you have Unity 2022.2 or higher. https://github.com/ScottKane/MSBuild4Unity-Sample/tree/grpc-lite

Once open in the Editor you will want to delete the Dependencies folder (from inside Unity), then on the top bar MSBuild -> Build All. This will regenerate the Dependencies folder, then go Build -> Build Server (Windows). You should then be able to play in the Editor which will launch the server too. For some reason you sometimes get a connection refused after an initial build, if this happens just stop the play session and hit play again (I have an open issue on aspnetcore about this as for some reason neither the appsettings.json or manual host filters are respecting AllowedHosts = *)

ScottKane:
Hmm, the aspnet guys seem to think the LiteServer is what's rejecting the connection (see https://github.com/dotnet/aspnetcore/issues/41171#issuecomment-1100416046). Any ideas?

ScottKane:
Ok so it looks like the connection refusal problem is caused by the client trying to connect before the server is actually up.

ScottKane:
Temporary fix for now:

```c#
public async Task Connect()
{
    while (true)
    {
        try
        {
            _channel = await ConnectionFactory
                .ConnectSocket(new IPEndPoint(IPAddress.Loopback, 5000))
                .AsStream()
                .AsFrames()
                .CreateChannelAsync();
            break;
        }
        catch (Exception e)
        {
            // Ignore SocketException when server hasn't started yet
        }
    }
}
```

Maybe allow for retries on the ConnectionFactory? The issue of the server only receiving the first message is still happening though and the client never gets a response from the server.

ScottKane:
You should just be able to clone and run https://github.com/ScottKane/MSBuild4Unity-Sample/tree/grpc-lite and see this behaviour.

PR #263, opened 2022-11-21
Bump minimum protobuf-net version

bdovaz:
- Use protobuf-net 3.x
- Add protobuf-net.ServiceModel in example projects as this functionality is now in a separate package

Reasons for this change:
- Unity only works with .NET Standard 2.0 and protobuf-net version 2.x carries over dependencies from .NET Standard 1.1 such as System.Reflection.Emit 4.3.0 and the like.
- In order to add it to this registry, it needs to meet this requirement: https://github.com/xoofx/UnityNuGet
- I understand that since 2 years have passed since the release of protobuf-net 3.x there is no impediment for this dependency to be on the 3.x branch instead of 2.x.

### DIFF
--- a/Directory.Build.props
+++ b/Directory.Build.props
@@ -33,6 +33,7 @@
 
     <ProtoBufNet2Version>2.4.6</ProtoBufNet2Version>
     <ProtoBufNet3Version>3.0.101</ProtoBufNet3Version>
+    <ProtoBufNetServiceModelVersion>3.0.101</ProtoBufNetServiceModelVersion>
   </PropertyGroup>
   <PropertyGroup Condition="'$(Configuration)'=='Release' or '$(Configuration)'=='VS'">
     <GenerateDocumentationFile>true</GenerateDocumentationFile>
--- a/examples/perf/PerfClient/PerfClient.csproj
+++ b/examples/perf/PerfClient/PerfClient.csproj
@@ -17,6 +17,7 @@
       <IncludeAssets>runtime; build; native; contentfiles; analyzers; buildtransitive</IncludeAssets>
     </PackageReference>
     <PackageReference Include="Grpc.Net.Client" Version="$(GrpcDotNetVersion)" />
+    <PackageReference Include="protobuf-net.ServiceModel" Version="$(ProtoBufNetServiceModelVersion)" />
   </ItemGroup>
   <ItemGroup>
     <ProjectReference Include="..\..\..\src\protobuf-net.Grpc.Native\protobuf-net.Grpc.Native.csproj" />
--- a/examples/perf/PerfServer/PerfServer.csproj
+++ b/examples/perf/PerfServer/PerfServer.csproj
@@ -16,6 +16,7 @@
       <PrivateAssets>all</PrivateAssets>
       <IncludeAssets>runtime; build; native; contentfiles; analyzers; buildtransitive</IncludeAssets>
     </PackageReference>
+    <PackageReference Include="protobuf-net.ServiceModel" Version="$(ProtoBufNetServiceModelVersion)" />
   </ItemGroup>
   <ItemGroup>
     <ProjectReference Include="..\..\..\src\protobuf-net.Grpc.AspNetCore\protobuf-net.Grpc.AspNetCore.csproj" />
--- a/src/protobuf-net.Grpc/protobuf-net.Grpc.csproj
+++ b/src/protobuf-net.Grpc/protobuf-net.Grpc.csproj
@@ -7,7 +7,10 @@
   </PropertyGroup>
   <ItemGroup>
     <PackageReference Include="Grpc.Core.Api" Version="$(GrpcVersion)" />
-    <PackageReference Include="protobuf-net" Version="$(ProtoBufNet2Version)" />
+    <PackageReference Include="protobuf-net" Version="$(ProtoBufNet3Version)" />
+  </ItemGroup>
+  <ItemGroup Condition="'$(TargetFramework)' == 'net461'">
+    <PackageReference Include="System.ValueTuple" Version="4.5.0" />    
   </ItemGroup>
   <ItemGroup Condition="'$(TargetFramework)' == 'net461' or '$(TargetFramework)' == 'netstandard2.0'">
     <PackageReference Include="Microsoft.Bcl.AsyncInterfaces" Version="5.0.0" />
--- a/tests/protobuf-net.Grpc.Test/protobuf-net.Grpc.Test.csproj
+++ b/tests/protobuf-net.Grpc.Test/protobuf-net.Grpc.Test.csproj
@@ -21,6 +21,7 @@
     </PackageReference>
     <ProjectReference Include="..\..\src\protobuf-net.Grpc\protobuf-net.Grpc.csproj" />
     <PackageReference Include="Grpc.Core.Api" Version="$(GrpcVersion)" />
+    <PackageReference Include="protobuf-net.ServiceModel" Version="$(ProtoBufNetServiceModelVersion)" />
   </ItemGroup>
 
   <ItemGroup Condition="'$(TargetFramework)' != 'net461' and '$(TargetFramework)' != 'net472'">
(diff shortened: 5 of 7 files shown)


mgravell:
>I understand that since 2 years have passed since the release of protobuf-net 3.x there is no impediment for this dependency to be on the 3.x branch instead of 2.x.

There are reasons for this decision. The consumer can update protobuf-net to V3 locally, which should have the desired result. The V2 lib multi-targets, and the netstandard bits should vary per target.

I think we need to be very clear as to what the objectives are here, rather than jumping straight it with code changes.

bdovaz:
> I think we need to be very clear as to what the objectives are here, rather than jumping straight it with code changes.

As I said:

"Unity only works with .NET Standard 2.0 and protobuf-net version 2.x carries over dependencies from .NET Standard 1.1 such as System.Reflection.Emit 4.3.0 and the like."

NickCraver:
@bdovaz You can use protobuf-net 3.x already, nothing is stopping you. NuGet dependencies are _minimum_ versions, nothing stops you from referencing a newer version of any downstream dependency unless there's a breaking change that prevents it.  You do this by referencing protobuf-net 3.x yourself, directly, rather than depending on the transitive dependency. 

What this PR does is make the decision for everyone and may break some other scenarios, when your goal isn't actually prevented today. For NuGet dependencies you always want as wide of an acceptable version range as possible - some people are on protobuf-net 2.x and need to be, this would prevent them from using the library going forward when we need not.

mgravell:
Likewise, protobuf-net v3, when targeting netstandard 2.0, still has those
dependencies you are opposed to - albeit at a slightly higher version (see
https://www.nuget.org/packages/protobuf-net#dependencies-body-tab) - but
you should be able to bump those versions locally. My hope is that if I can
complete the AOT work we can lose those dependencies completely, especially
for scenarios like Unity, but: that's going to need quite some more work
yet.

NickCraver:
@bdovaz I'm not sure we're getting the correct info across here. The dependencies are *minimum* versions. You are absolutely free to reference a *higher* version, without forcing anyone else to.

Put another way: what stops you from referencing protobuf-net 3.x directly in your project? What breaks?

bdovaz:
I think you are not understanding me, have you looked at @xoofx's [UnityNuGet project](https://github.com/xoofx/UnityNuGet)?

As I said, that project has a [concrete dependency listing](https://github.com/xoofx/UnityNuGet/blob/master/registry.json) whose minimum versions are those with .NET Standard 2.0+ compatibility.

That's why we can't add protobuf-net.Grpc, because they depend on protobuf-net 2.x and this in turn has dependencies like System.Reflection.Emit in 4.3.0 (it doesn't target .NET Standard 2.0).

The solution as I said is to release a 2.4.8 version of protobuf-net 2.x with those dependencies lifted which I don't think is that much of a problem: https://www.nuget.org/packages/protobuf-net/2.4.7#dependencies-body-tab

I can create a PR in protobuf-net that will bump those dependencies so you can release a 2.4.8 in NuGet and then another PR in this repository lifting the dependency from 2.4.7 to 2.4.8 which I don't think will be much of a problem either.

NickCraver:
>That's why we can't add protobuf-net.Grpc, because they depend on protobuf-net 2.x and this in turn has dependencies like System.Reflection.Emit in 4.3.0 (it doesn't target .NET Standard 2.0).

This is factually incorrect. It depends on protobuf-net >= 2.x. Importantly: **greater than or equal to**, that's how NuGet dependencies work and what I'm trying to convey here. Think about what would happen if this _wasn't_ the case, every vulnerability in any downstream dependency would require a new build of everyone that uses it to get a fix out, that would be pure madness.

You simply reference __both__ libraries here. `protobuf-net.Grpc` and the version of `protobuf-net` you want, which can be 3.x.

bdovaz:
@NickCraver you still don't understand me, the problem is not in NuGet but in Unity + UnityNuGet + the dependencies of these packages that I mention from protobuf-net. The way UnityNuGet works, it requires that the listing defined in the registry.json file that I have given you has all the dependencies chain with the minimum version compatible with .NET Standard 2.0+.

In that file you have for example the System.Reflection.Emit package (on which protobuf-net depends) with the minimum version 4.6.0 defined because it is the first one that works in Unity since it targets .NET Standard 2.0. For this same reason we can no longer add the protobuf-net package to this registry, because the 4.3.0 version does not exist since we have set the 4.6.0 (required by Unity).

Finally, so much problem is to raise dependencies type System.Reflection.Emit that 4.3.0 is 6 years old? The 4.6.0 is 3 years old and it is compatible with .NET Standard 2.0 so it solves the problem.

Unity as you know is a multiplatform game engine used by millions of people so making it compatible is very important.

xoofx:
> It depends on protobuf-net >= 2.x
> because the 4.3.0 version does not exist since we have set the 4.6.0 

@bdovaz Couldn't this be solved by implementing in UnityNuGet a >= resolution for dependency as long as e.g we stay within a minor version (for implicit resolution)? PR would be welcome.

bdovaz:
> > It depends on protobuf-net >= 2.x
> > because the 4.3.0 version does not exist since we have set the 4.6.0
> 
> @bdovaz Couldn't this be solved by implementing in UnityNuGet a >= resolution for dependency as long as e.g we stay within a minor version (for implicit resolution)? PR would be welcome.

Could you explain it better with an example for this case we are talking about? So I can understand it better and think if I could get to do it.

**I also still think** that even if we can do the PR that @xoofx mentions in UnityNuGet, I don't think it's that much of a problem to lift 6 year old dependencies to 3 year old ones in protobuf-net package.

NickCraver:
> I also still think that even if we can do the PR that @xoofx mentions in UnityNuGet, I don't think it's that much of a problem to lift 6 year old dependencies to 3 year old ones in protobuf-net package.

That's not what this PR does though, this PR is asking that everyone who uses protobuf-net.Grpc to _require_ protobuf-net 3.x, which isn't fully feature compatible with 2.x, some functionality was dropped and some people are intentionally staying on 2.x for those features. Even if we only changed the dependency versions on protobuf-net 2.x, there's still friction though due to strong naming. You may not think it's much of a problem, but it's can easily create a major pain for many consumers of the package - we've dealt with this first hand many, many, many times.

bdovaz:
I know that this PR does not apply the changes in which this conversation has derived (raise the dependencies of protobuf-net to minimum versions compatible with .NET Standard 2.0), in fact if we were to do that PR would be in the protobuf-net repository not in this one (protobuf-net.Grpc).

I still do not understand how it is such a big problem to create a PR with the minimal changes that I say as changing the protobuf-net dependency of System.Reflection.Emit from 4.3.0 to 4.6.0. Following the creation of UnityNuGet over the years I have encountered packages with the same problem as protobuf-net and talking to the mantainers all have understood the problem and have accepted the PR without problem to make it compatible with Unity.

If in this case you consider it such a big problem as not to accept the PR that I mention that could be created, I'm afraid that we will not be able to make it compatible with Unity through UnityNuGet.

xoofx:
> If in this case you consider it such a big problem as not to accept the PR that I mention that could be created, I'm afraid that we will not be able to make it compatible with Unity through UnityNuGet.

I think that can be workaround in UnityNuGet frankly. We can "hardcode" (even via the registry json file) a kind of binding redirect that says that when e.g you have a `4.3.0` dependency, than it is requested as `4.6.0`.

bdovaz:
@xoofx I'm going to close this PR, create an issue in UnityNuGet and take the conversation there because seeing where the conversation has gone it doesn't make sense to continue it here.

NickCraver:
> I still do not understand how it is such a big problem to create a PR with the minimal changes that I say as changing the protobuf-net dependency of System.Reflection.Emit from 4.3.0 to 4.6.0

This is due to strong naming, when you have 15x the downloads on the 4.3.0 version vs 4.6.0 that means most things have that dependency, by far. This means when a library beside ours wants 4.3.0 and we require 4.6.0 there now needs to exist a binding redirect in all downstream consumers for their application to run due to 4.3.0 not being there anymore. This creates a lot of friction for users to the point we _really_ try to avoid it. It's also why .NET Core+ removed this entire concept, because it's so painful. That's why it's so important to us to not bump dependencies unless we _have_ to, and in this case we don't have to, an alternative is on the table that supports all use cases with maximum compatibility.

PR #275, opened 2023-02-09
Update ServicesExtensions.cs

pingvin1308:
replace string format by index with names

it can be useful for log indexing systems that tries to index values that services passes into the string.

### DIFF
--- a/src/protobuf-net.Grpc.AspNetCore/ServicesExtensions.cs
+++ b/src/protobuf-net.Grpc.AspNetCore/ServicesExtensions.cs
@@ -43,7 +43,7 @@ namespace ProtoBuf.Grpc.Server
             void IServiceMethodProvider<TService>.OnServiceMethodDiscovery(ServiceMethodProviderContext<TService> context)
             {
                 int count = new Binder(_logger).Bind<TService>(context, _binderConfiguration);
-                if (count != 0) _logger.Log(LogLevel.Information, "RPC services being provided by {0}: {1}", typeof(TService), count);
+                if (count != 0) _logger.Log(LogLevel.Information, "RPC services being provided by {service}: {count}", typeof(TService), count);
             }
         }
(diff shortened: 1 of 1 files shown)


mgravell:
Fair enough; found the docs: https://learn.microsoft.com/en-us/aspnet/core/fundamentals/logging/?view=aspnetcore-7.0#log-message-template - checks out and makes sense

mgravell:
can you update /docs/releasenotes.md too, please? see existing content for format - this would be in the "unreleased" section currently

mgravell (approved):

src/protobuf-net.Grpc.AspNetCore/ServicesExtensions.cs
```diff
@@ -43,7 +43,7 @@ public CodeFirstServiceMethodProvider(ILoggerFactory loggerFactory, BinderConfig
             void IServiceMethodProvider<TService>.OnServiceMethodDiscovery(ServiceMethodProviderContext<TService> context)
             {
                 int count = new Binder(_logger).Bind<TService>(context, _binderConfiguration);
-                if (count != 0) _logger.Log(LogLevel.Information, "RPC services being provided by {0}: {1}", typeof(TService), count);
+                if (count != 0) _logger.Log(LogLevel.Information, "RPC services being provided by {service}: {count}", typeof(TService), count);
```
meirkr, line 46: Hi
The convention is to have PascalCase:
[CA1727: Use PascalCase for named placeholders](https://learn.microsoft.com/en-us/dotnet/fundamentals/code-analysis/quality-rules/ca1727)
```suggestion
                if (count != 0) _logger.Log(LogLevel.Information, "RPC services being provided by {Service}: {Count}", typeof(TService), count);
```

mgravell:
merging, thanks

PR #331, opened 2024-07-04
Improve error message when binding methods

BasConijn:
I am converting some WCF services to GRPC using this (awesome) package. However sometimes there are methods with overloads in the service class. This throws an "An item with the same key has already been added", though the message did not mention the name of the method it was trying to bind.


mgravell (approved):
LGTM, thanks! For the record, can you confirm that you're able and willing to contribute this to the project in line with the project license?

BasConijn:
Yes, i am willing to contribute.

PR #355, opened 2025-04-29
Add SchemaGenerationOptions overwrite to SchemaGenerator.

DxsSucuk:
This PR allows devs to overwrite the default values via the property Options and customize stuff like the generation flags.

### DIFF
--- a/src/protobuf-net.Grpc.Reflection/SchemaGenerator.cs
+++ b/src/protobuf-net.Grpc.Reflection/SchemaGenerator.cs
@@ -17,7 +17,15 @@ namespace ProtoBuf.Grpc.Reflection
         /// <summary>
         /// Gets or sets the syntax version
         /// </summary>
-        public ProtoSyntax ProtoSyntax { get; set; } = ProtoSyntax.Proto3;
+        public ProtoSyntax ProtoSyntax { get => Options.Syntax; set => Options.Syntax = value; }
+
+        /// <summary>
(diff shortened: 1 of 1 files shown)


mgravell:
Thanks for this, and apologies for the very long silence — the idea is right and it should have been picked up much sooner.

I've pushed two small follow-ups directly to the branch (so heads-up: your fork's `main` has moved) rather than send it back for another round after this long:

- **`Options` is now get-only.** `ProtoSyntax` reads through it, so a settable property meant `Options = null` turned a plain property access into a `NullReferenceException`. It's configured in place instead, which is also how `Services` and `Types` on the options themselves already work.
- **`Types` is copied onto the per-generation instance.** Exposing `Options` makes `generator.Options.Types.Add(...)` look supported, and without this it silently produced nothing — the only member that did.

The doc comment now describes what the property *is* — a template, copied onto a fresh instance per generation, including `Types` but not `Services` or `Package`, which are derived from the contract types being generated. That last part is worth stating, since setting `Options.Services` would otherwise look like it should work.

Builds clean and the Reflection tests pass. Merging from here — thanks again.

mgravell:
Sorry for the delay - I was changing jobs, and: I'm human. So: merging *with some tweaks* - we don't need this to be settable, and it opens in NRE issues; I'm making it get-only - still editable in place so fully configurable. Thanks, good contribution.

PR #359, opened 2025-11-05
Fix - Add support for multiple AssemblyLoadContexts

derstpal:
Hello,

We are using a Plugin pattern for our dotnet application. Each plugin can use its own GRPC contract in isolation. The contract is loaded inside an AssemblyLoadContext. Plugins also share some common dependencies like protobuf-net.Grpc

When we Create a proxy on a same contract but loaded from different ALCs, we have various Reflection Type exceptions.
This PR contains a fix for this case and remain compatible on one default ALC and also other targetted frameworks.

Some docs:
[Contextual Reflection](https://github.com/dotnet/coreclr/blob/v3.0.0/Documentation/design-docs/AssemblyLoadContext.ContextualReflection.md)


mgravell:
Thanks for this, and apologies for the long silence — the scenario is legitimate and the PR is a real fix.

I've rebased the branch onto current `main` and added fixup commits, so it's ready to go. **Note this was a force-push** (a rebase rewrites SHAs); your three commits are unchanged in content and still attributed to you, they just have new hashes.

### Why the rebase was needed

The PR predates a few things: the solution moved to `.slnx`, and the test project moved to `net472;net8.0;net10.0` on xunit.v3. The only conflict was that churn — main's version stands, plus your one genuine addition (the `TestProxy` project reference).

### The fixups

- **`protobuf-net.Grpc.Tests.TestProxy` moved `src/` → `tests/`**, and marked `IsPackable=false`. This is the one that mattered: `src/**` *is* the ship list — `Build.csproj` uses exactly that glob when `Packing=true` — so a test fixture there would have been published to nuget.org as a seventh package. Verified `pack` still produces exactly the same six.
- **`IsTestProject=false`** on that project, which is what was breaking CI. `Directory.Build.props` infers test-ness from the project *name* containing "Test", so `dotnet test` on the traversal tried to run VSTest against a fixture library with no adapter and aborted. Not something the move introduced — it would have failed the same way in the original layout.
- **Retargeted `net6.0;net8.0` → `net8.0;net10.0`**, matching the suite that consumes it; net6.0 is out of support.
- **Solution entry into `.slnx`** (the rebase had landed the original `.sln` edit on `protobuf-net.Grpc.sln.old`, which is a stale leftover).
- **A comment** on `_proxyModules` explaining why strong ALC keys are safe — a collectible ALC only unloads on an explicit `Unload()`, and the `Unloading` handler removes it before collection. It reads like a leak at a glance and deserved a line saying why it isn't.
- **`#if NET`** in place of `#if NET6_0_OR_GREATER` for the four gates this PR adds; `NET` is structurally identical to `NET5_0_OR_GREATER` and says the intended thing.

### On the fix itself

I checked the part I couldn't take on faith — whether `EnterContextualReflection()` really governs which ALC a *dynamically defined* assembly lands in, as opposed to just `Assembly.Load` resolution. Your tests answer it: reverting only `ProxyEmitter`/`ProxyModuleHelper` and leaving everything else gives

```
Object of type 'Grpc.Core.Marshaller`1[...TestProxy.In]' cannot be converted to
type 'Grpc.Core.Marshaller`1[...TestProxy.In]'
```

— the same type from two load contexts, which is exactly the symptom. Tests that demonstrably fail without the change are the most valuable thing in a PR like this, so thank you for including them.

Full solution builds; 12 test runs green in both Release and Debug.

mgravell:
One more commit: the four conditional-compilation gates this PR introduces now use `#if NET` rather than `#if NET6_0_OR_GREATER`.

Equivalent here — verified per-TFM rather than assumed; `NET`, `NET6_0_OR_GREATER` and `NETCOREAPP3_0_OR_GREATER` are all defined on exactly `net8.0` and `net10.0`, and on none of `net472`, `netstandard2.0` or `netstandard2.1`. But `#if NET` says the intended thing — *modern .NET, as opposed to netfx or netstandard* — where `NET6_0_OR_GREATER` implies a version floor that isn't real, and reads as though net5.0 were excluded deliberately when it just isn't in the TFM list. It also can't go stale if a TFM is added.

Scoped to this PR's own gates; the rest of the repo is handled separately in #371.

Still green in both Release and Debug.

mgravell (approved):

PR #364, opened 2026-05-21
Source-generate gRPC proxies + server bindings for AOT/trim (closes #79)

### LINKED ISSUE #79: Use static proxy generator instead of IL emitting?
This package implements interface proxy by IL emitting which is not available when AOT compilation is used.
Consider generating the proxy classes on build.
See:
[Microsoft.Orleans.CodeGenerator.MSBuild](https://www.nuget.org/packages/Microsoft.Orleans.CodeGenerator.MSBuild/)
[Fody](https://github.com/Fody/Fody)

virzak:
Closes #79

## Summary

Adds a Roslyn source generator (`protobuf-net.Grpc.BuildTools`) that emits build-time client proxies AND server bindings for every interface tagged with `[Service]` or `[ServiceContract]` — no source changes to the consuming code, no opt-in attribute, no `partial` keyword. Registration happens through a `[ModuleInitializer]`-populated registry, so consumption is fully static: trimming and `PublishAot=true` shake the IL-emit / `Expression.Compile` fallback away.

## Consumer usage

```csharp
[Service]
public interface IGreeter
{
    ValueTask<HelloReply> SayHelloAsync(HelloRequest req, CallContext ctx = default);
}

// nothing else changes
var greeter = channel.CreateGrpcService<IGreeter>();
services.AddCodeFirstGrpc();
endpoints.MapGrpcService<MyGreeterImpl>();
```

The generator ships **inside** `protobuf-net.Grpc.nupkg` (under `analyzers/dotnet/cs/`), so a project that already references `protobuf-net.Grpc` (directly or transitively via `protobuf-net.Grpc.AspNetCore` / `.ClientFactory`) gets the generator with no extra package reference. It runs on every `[Service]`/`[ServiceContract]` interface in the consuming compilation; emits a client proxy + server bindings + a `[ModuleInitializer]` that registers both. Client and server projects pick them up transparently.

### Generator (`src/protobuf-net.Grpc.BuildTools/`)
- New analyzer-style project, targets `netstandard2.0`, packed under `analyzers/dotnet/cs/` inside `protobuf-net.Grpc.nupkg` (no separate package).
- `IIncrementalGenerator` scans `[Service]` and `[ServiceContract]` interfaces directly via `ForAttributeWithMetadataName`.
- Emits one `.g.cs` per interface containing:
  - `{Contract}_ClientProxy : ClientBase, IContract` — sealed proxy with explicit interface impls dispatching through `Reshape.*`, plus `public static IContract Create(CallInvoker, BinderConfiguration)` and `Create(CallInvoker)` factories. Marshallers and `Method<,>` are instance readonly fields built from the supplied configuration in the constructor — custom `MarshallerFactory` configurations apply the same as with the IL emitter.
  - `{Contract}_ServerBindings` — static class with `public static int Bind<TService>(IServerMethodBinder<TService> binder)` that registers one handler per operation. Each per-op block runs inside its own try/catch that **fail-fasts**: any per-method failure (no marshaller, custom factory rejects a type, etc.) is rethrown as an `InvalidOperationException` wrapping the original with the operation name + contract `FullName` in the message, so the host's startup log points straight at the broken method.
  - `{Contract}_Init` — `[ModuleInitializer]` calling `GeneratedProxyRegistry.RegisterClient<IContract>(...)` and `RegisterServer(typeof(IContract), typeof({Contract}_ServerBindings))`.
- Covers Unary, Server-streaming, Client-streaming, Duplex × `NoContext` / `CallContext` / `CancellationToken` with `IAsyncEnumerable<T>`. Interfaces with any method shape the generator doesn't handle (`IObservable<T>`, `Stream`, raw `CallOptions`/`ServerCallContext`, sub-services) are skipped entirely so the runtime IL/`Expression.Compile` fallback takes them over — zero behavior regression for existing consumers.
- Diagnostics `PBNG002`–`PBNG004` for nested, unsupported-shape, or generic interfaces.

### Runtime (`src/protobuf-net.Grpc/`)
- New `ProtoBuf.Grpc.Internal.GeneratedProxyRegistry` — public static (the `[ModuleInitializer]`s call into it), `[Browsable(false)]`/`[EditorBrowsable(Never)]`; concurrent dictionaries keyed on contract type.
- `ProxyEmitter.CreateFactory<TService>` lookup order: (1) `GeneratedProxyRegistry`, (2) `[Proxy(typeof(X))]` manual override, (3) IL emit fallback. The IL path is gated on `RuntimeFeature.IsDynamicCodeSupported` and throws a clear `NotSupportedException` pointing at the BuildTools package when dynamic code isn't available.
- `ServerBinder.Bind` lookup order: (1) `GeneratedProxyRegistry`, (2) the existing reflection + `Expression.Compile` path. The fallback's hot path (`MethodStub<TService>.CreateDelegate` and `AddMethod<,,>`) is annotated `[RequiresDynamicCode]` / `[RequiresUnreferencedCode]`.
- New `IServerMethodBinder<TService>` abstraction (+ four matching `TService`-generic handler delegates: `UnaryServerHandler<TService,TRequest,TResponse>` etc.) so generated code can register methods without a `Grpc.AspNetCore.Server` dependency. Per-op failures always throw (no opt-in tolerance) — the host startup fails loudly with operation context rather than silently skipping a broken method.
- New `IGeneratedServerBindContext` — the bind-state contract implemented by callers that can close the generic invocation against the generated `Bind<TService>` (AspNetCore's `OnServiceMethodDiscovery<TService>` qualifies).
- `ProxyEmitter`'s `s_module` (the `AssemblyBuilder.DefineDynamicAssembly` call) moved out of the static constructor into a lazy initialiser so the trimmer can drop the entire IL-emit codepath when no service needs it.
- `GoogleProtobufMarshallerFactory` rewritten without `Expression.Lambda(...).Compile()` — delegates are bound via `Delegate.CreateDelegate` against the parser + the static `MessageExtensions.WriteTo` / `ToByteArray` helpers.
- `ContractOperation`, `MarshallerCache`, `AttributeHelper`, and `ProxyAttribute` carry the `[RequiresDynamicCode]` / `[RequiresUnreferencedCode]` / `[DynamicallyAccessedMembers]` / `[UnconditionalSuppressMessage]` annotations needed for the trimmer to propagate cleanly through the gated IL-emit chain without leaking warnings onto the static path.

### AspNetCore (`src/protobuf-net.Grpc.AspNetCore/`)
- `CodeFirstServiceMethodProvider<TService>.OnServiceMethodDiscovery` now wraps the gRPC `ServiceMethodProviderContext<TService>` in a `GeneratedBindAdapter<TService>` that doubles as the bind state, `IServerMethodBinder<TService>`, and `IGeneratedServerBindContext`. When the registry has bindings for the contract, the adapter invokes the generated `Bind<TService>` (closing the generic with `MakeGenericMethod(typeof(TService))` exactly once per service registration); each generated handler then registers via `context.AddUnaryMethod` / etc.
- Falls back to the existing `Expression.Compile` path when no generated bindings exist.

## Verification

- `dotnet test`: 197 / 197 pass on `net10.0`. Includes generator smoke tests that drive the generated `Bind<MyService>` end-to-end through a fake `IServerMethodBinder<>` and verify (a) the handler dispatches through the interface, and (b) a marshaller-resolution failure surfaces as `InvalidOperationException` carrying the operation name + contract `FullName` + original exception as `InnerException`.
- `dotnet publish -p:PublishAot=true` on a probe console app: **zero** IL2026 / IL3050 / IL2070 / IL2075 / IL2087 / IL2090 / IL2060 warnings from `protobuf-net.Grpc.dll`.
- Multi-target build clean across `net472`, `netstandard2.0`, `netstandard2.1`, `net8.0`, `net10.0`.
- Validated end-to-end in a real consumer service: `dotnet publish -c release -a x64` with `PublishTrimmed=true`, all gRPC methods bind, host comes up cleanly.

## Not covered (follow-up candidates)

The generator skips and lets the runtime IL emitter / `Expression.Compile` take over (still functionally correct, not AOT-friendly) for these interface shapes:

- `IObservable<T>` for streaming arguments / returns.
- `Stream` / `Task<Stream>` / `ValueTask<Stream>` for byte-streaming server methods.
- Sub-services (`[SubService]`).
- Generic interfaces.
- Raw-gRPC patterns (`CallOptions`, `ServerCallContext` on client interfaces) — but the IL emitter throws on these too, so behavior is unchanged.

Generation is all-or-nothing per interface; if a contract mixes supported and unsupported methods, none of its methods take the generated path. This deliberately avoids any regression and gives a clean migration story (split the contract, or open an issue requesting the missing shape).

## Important caveat for trim/AOT consumers

`protobuf-net.Grpc.dll` is now AOT-clean, but `protobuf-net.dll` itself is still reflection-based — `RuntimeTypeModel.Default.CanSerialize(typeof(T))` walks `T`'s members at runtime to discover `[ProtoContract]` / `[ProtoMember]`. Under `PublishTrimmed=true`, consumers must root the assemblies whose contract types they want to serialize:

```xml
<ItemGroup>
  <TrimmerRootAssembly Include="MyApp.Contracts" />
  <TrimmerRootAssembly Include="protobuf-net" />
  <TrimmerRootAssembly Include="protobuf-net.Core" />
</ItemGroup>
```

Under `PublishAot=true` protobuf-net's reflection breaks more fundamentally and the only path forward is a generator-based serializer (Google.Protobuf, MemoryPack, MessagePack-CSharp). This PR doesn't try to solve that — it makes the gRPC plumbing layer AOT-clean, the serializer layer is independent.

## Test plan

- [x] Existing tests pass
- [x] New generator smoke tests pass (client proxy returned from registry, server `Bind<T>` dispatches through interface, bind failure surfaces with operation context + original exception preserved as `InnerException`)
- [x] `PublishAot=true` produces zero IL trim/AOT warnings from `protobuf-net.Grpc.dll`
- [x] Real-world publish of an AspNetCore service with `PublishTrimmed=true`: all methods bind, host runs

🤖 Generated with [Claude Code](https://claude.com/claude-code)


virzak:
Hi, this is @virzak not AI,

Feeling a bit weird creating a PR this way. It definitely did a good job on my project which consumed the end result, but as with most AI generated commits, the biggest challenge is finding time to understand 100% of what it did. I hope that the end result of this effort is productive, and doesn't create a ton of work for the maintainer(s).

Please let me know your thoughts and I'll be happy to share the burden of diving deep into this. Also this PR can be split into multiple smaller PRs.

mgravell:
Yep, sounds interesting, and is something I have planned for ages but not found time for. I'll take a look when I get a moment. I wonder whether this might locate better inside protobuf-net.BuildTools, but that's not a huge issue.

virzak:
Happy to move parts of this PR to protobuf-net.BuildTools if you prefer. It would be a cross repo move and it would have to be published there first, so that this repo could consume it.

mgravell:
OK, this is amazing work, thanks. I want to propose we move it over to the existing protobuf-net.BuildTools codebase, which already handles protobuf-net.Grpc, but I don't want to steal your credit. Are you happy to resubmit there? Are you happy for me to simply pick it up and move it (which would lose credit, except for as text)? Or?

For additional context / timing: https://github.com/protobuf-net/protobuf-net/pull/1254

virzak:
Happy to resubmit - done: protobuf-net/protobuf-net#1255 moves the generator into `protobuf-net.BuildTools`, next to `ServiceContractAnalyzer` and the `[ProtoModel]` work in protobuf-net/protobuf-net#1254.

This PR is now the runtime half only: `GeneratedProxyRegistry`, `IServerMethodBinder<T>`, and the client-factory / `ServerBinder` hooks that consult them before falling back to ref-emit. It drops from ~1,900 to ~850 added lines.

Splitting the two across packages did need some new work, since the generator can now be a different version than the runtime it emits calls into. It stands down silently unless it can see `GeneratedProxyRegistry` / `IServerMethodBinder<>` / `Grpc.Core.ClientBase` in the compilation - which also means it no longer fires on a plain WCF `[ServiceContract]` in a project that doesn't use protobuf-net.Grpc at all - and reports PBN3000 / PBN3004 when the language version or target framework can't support the emitted code. Ordering-wise this side is safe to merge first or second, but runtime-first is tidier.

Porting it also turned up three bugs that were in this PR as it stood:

- explicit interface implementations of inherited operations were qualified with the derived interface instead of the declaring one, so the generated proxy didn't compile;
- inherited interfaces were bound wholesale, where the runtime only binds `[SubService]` bases - that would have put operations on the wire that the runtime never binds;
- `IObservable<T>`, `Stream` and `Grpc.Core`'s own call types were being mistaken for ordinary payloads, which compiles and then fails looking for a marshaller at startup.

The first one was caught by the new tests actually compiling the generated code, which felt worth the setup cost.

`GenerateProxyTests` here becomes `GeneratedProxyRegistryTests`: it can't run the generator any more, so it hand-writes what the generator emits and pins the contract that code depends on. The emitted text is pinned by golden files on the other side.

One unrelated observation while running the suite: `StreamTests.DuplexEchoObservable` is timing-dependent - the producer is a hot `Subject` that emits its first item after 10ms, so if the client hasn't subscribed by then item 0 is simply lost. On my machine it fails maybe half the time in one checkout and not at all in another, on `main` as much as on this branch. Might be worth a `Replay`/`ReplaySubject` or an explicit subscription handshake, but I've left it alone here.

mgravell:
See https://github.com/protobuf-net/protobuf-net/pull/1255#issuecomment-5307850366

mgravell:
Closing after discussion with contributor; intent is captured and agreed. Cross-ref: https://github.com/protobuf-net/protobuf-net/pull/1282

mgravell:
So: we have a thing! please see https://grpc.protobuf-net.dev/aot and https://docs.protobuf-net.dev/aot which now work together, in theory; new release on way to NuGet

Feedback very welcome! Much of your work remains in spirit, even if we implemented bits in different ways. I think the chosen direction makes sense, at least.

PR #365, opened 2026-08-16
AOT: preview trigger attributes, and stop rooting RuntimeTypeModel where ref-emit is absent

mgravell:
Draft. The protobuf-net.Grpc half of build-time gRPC proxies — see
protobuf-net/protobuf-net#1282 for the generator, and
protobuf-net/protobuf-net.Grpc#364 / protobuf-net/protobuf-net#1255 by @virzak for the work this
builds on.

Three changes.

### `[ProtoGrpc]` / `[ProtoService]`

The trigger attributes for the BuildTools generator. `[Experimental]` with protobuf-net's own
**PBN9001** id, so one `NoWarn` covers both halves of the AOT story; `ExperimentalAttribute` is
polyfilled below net8.0, mirroring protobuf-net.Core.

These are currently generator-owned (emitted `internal` per consuming assembly), which is how
`[ProtoModel]` started. Moving them here is not a break in either direction: an `internal` copy in
the consumer's own assembly wins name resolution over a `public` one in a referenced assembly.

### Don't root `RuntimeTypeModel.Default` where ref-emit is absent

`ProtoBufMarshallerFactory.Default` and `.Create` no longer name `RuntimeTypeModel.Default` when
`RuntimeFeature.IsDynamicCodeSupported` is false. The gate rather than a `Requires*` annotation is
deliberate: ILC substitutes it for a constant and eliminates the arm *before* trim analysis, so the
demand disappears instead of relocating to callers — the same lever protobuf-net's
`TypeModel.ResolveSerializer<T>` uses, and that #364 applies to `ProxyEmitter`.

**Deliberate behaviour change:** under native AOT, a consumer with no compile-time model now gets
"no marshaller available" by name, rather than the same failure arriving via a swallowed
`MakeGenericType` inside `DynamicStub`. Neither can work; this one says so.

### `BinderConfiguration.Binder` is public

While it was `internal`, generated bindings had no way to reach the configured binder and fell back
to `ServiceBinder.Default` — silently ignoring a consumer's override. This is a bug fix independent
of AOT.

## Measurement

Native `win-x64` publish of protobuf-net's `src/AotGrpcSmoke` — a real client and server over a real
socket, with generated serializers and generated proxies:

| | IL warnings | bytes |
| --- | ---: | ---: |
| baseline (shipped 1.2.2) | 100 | 15,015,936 |
| marshaller gates only (this PR) | **100** | 15,019,520 |
| `BytesValue.SlowParse` gated only | **100** | 15,017,984 |
| both | **5** | 14,469,632 |

**The two roots are an AND, not an OR.** Either alone buys exactly nothing, because the other keeps
the whole reflective serializer machinery reachable. Worth recording, because measured in isolation
each of these changes looks worthless — and the natural conclusion would have been to drop them.

Together: **100 → 5 warnings, −549,888 bytes (−3.7%)**. Of the residual 5, four are per-assembly
`IL2104`/`IL3053` rollups; the one real warning is an `IL2091` in protobuf-net's *generated* code and
is fixed on that side.

## Still outstanding: `BytesValue.SlowParse`

Deliberately **not** included here. `BytesValue.cs:186` does
`IProtoInput<Stream> model = RuntimeTypeModel.Default;`, and because `MarshallerCache`'s field
initialiser pre-registers `BytesValue.Marshaller`, *every* cache roots it — which is why it dominates
the measurement above.

It cannot simply be gated: it is a real fallback for payloads `TryFastParse` declines (unknown
fields, length prefixes over three bytes), so throwing there would break genuine traffic. It wants a
hand-written reader instead, and that touches the careful fast-path code — happy to do it, but it
merits a look rather than a guess, particularly around merge semantics for a repeated field 1
(protobuf-net's `AppendBytes` concatenates).

## Notes

- 194 tests pass on net8.0 and net10.0, 192 on net472.
- This repo is still on AppVeyor; a preview package would want the GitHub Actions + trusted
  publishing setup protobuf-net now uses (`.github/workflows/dotnet.yml` + `release.yml` there as a
  template).


mgravell:
Merged `main` (the new Actions pipeline) and re-verified with the exact commands CI runs: build green, and all suites pass — 194 net8.0 / 194 net10.0 / 192 net472 for the main tests, plus Reflection (14 × 3) and Integration (54/53 × 4).

Marking ready. One thing deliberately still outstanding and called out in the description: `BytesValue.SlowParse` is the *other* half of the `RuntimeTypeModel` rooting, and without it these gates measure as buying nothing (100 → 100 warnings). It needs a hand-written reader rather than a gate, and there's a semantics question worth a decision first — what a repeated field 1 should do, given protobuf-net's `AppendBytes` concatenates where a naive hand-rolled "last wins" would silently differ.

mgravell (commented):
👍

PR #366, opened 2026-08-16
Release via GitHub Releases: trusted-publishing workflow replaces AppVeyor/MyGet

mgravell:
Ports protobuf-net's release pipeline (protobuf-net/protobuf-net#1258) to this repo, so a preview
package can be published for the build-time gRPC proxy work
(protobuf-net/protobuf-net#1282, #365).

## What it does

- **`dotnet.yml`** — build + test on push/PR to main, windows-latest, SDKs 8.0.x and 10.0.x.
- **`release.yml`** — a published GitHub Release builds, tests, packs the ship list and pushes to
  nuget.org via OIDC trusted publishing. No stored API key. `workflow_dispatch` runs everything
  except the tag check and the push, as a dry run, keeping the packages as a run artifact so the
  pipeline can be proven before a release is cut.

Matches the policy already configured on nuget.org:

| | |
| --- | --- |
| Repository Owner | `protobuf-net` |
| Repository | `protobuf-net.Grpc` |
| Workflow | `release.yml` |
| Environment | `release` |

**Neither the workflow file name nor the environment name is free to change** — both are part of that
policy, so there's a comment at the top of the file saying so.

## Verified locally

Not just written — run, with the exact commands the workflows use:

- `dotnet build Build.csproj -c Release /p:CI=true /p:GeneratePackageOnBuild=false` — succeeds;
- `dotnet test Build.csproj --no-build -c Release /p:CI=true` — 194 net8.0, 194 net10.0, 192 net472,
  plus the Reflection and Integration suites, all green;
- `dotnet pack Build.csproj --no-build -c Release /p:Packing=true` — produces all six packages
  (`protobuf-net.Grpc`, `.AspNetCore`, `.AspNetCore.Reflection`, `.ClientFactory`, `.Native`,
  `.Reflection`);
- `nbgv get-version --variable NuGetPackageVersion` from the repo root — `1.2.6-g4c541366d5`.

That last one is the one difference from protobuf-net worth knowing: **`version.json` lives at the
repository root here**, not under `src/`, so the guard step calls `nbgv` without `--project`. With
the wrong path it would silently compute a different version and fail every release tag.

## AppVeyor retires

Same reasoning as protobuf-net/protobuf-net#1258: a `Visual Studio 2019` image it can no longer
build on, and it was the only thing feeding MyGet. The readme badge moves to the Actions one.

Easy to revert if you'd rather keep MyGet publishing for now — it's one file plus a badge.

## Note

`NUGET_USER` needs to exist as a repository secret (holding the nuget.org profile name), and the
`release` environment needs to exist in repository settings for the environment protection to apply.

Unrelated and left alone: the readme's logo still points at the dead-ish `protogen.marcgravell.com`
host. protobuf-net's migration commit cleaned up its equivalent links; worth doing here too, but not
in a CI PR.


mgravell (commented):
👍

PR #369, opened 2026-08-16
Fixed custom attributes not collected with subservice interface (resolved #330)

### LINKED ISSUE #330: Missed custom attribute in SubService interface
We use custom attribute to decorate grpc service class methods in order to do something by wrapping the method implementation through asp net core middleware.
All custom attributes of a method of a controller (or of a method of the grpc service) will be available through aspnet core endpoint's metadata.

So, we have this scenario:
```
[SubService]
public interface ITestSubService 
{
   [Operation]
   ValueTask DoSomethingCommon(CallContext context);
}

[Service]
public interface ITestService : ITestSubService 
{
   [Operation]
   ValueTask DoSomethingSpecific(CallContext context);
}

internal class TestService : ITestService 
{
   [CustomAttribute]
   public ValueTask DoSomethingCommon(CallContext context) 
   {
      // ...
   }

   [CustomAttribute]
   public ValueTask DoSomethingSpecific(CallContext context) 
   {
      // ...
   }
}
```

And all worked good until today when we added the custom attribute also to the method inherited by the subservice interface `DoSomethingCommon` and we noticed that the attribute will never added to the endpoint's metadata collection.
All other services and methods which have the attributes works good.

By debugging we encontered the "issue" [here](https://github.com/protobuf-net/protobuf-net.Grpc/blob/main/src/protobuf-net.Grpc/Configuration/ServerBinder.cs#L67).
The `bindCtx` is created in the foreach loop of each type which methods have to be bound (so both interfaces), but the parameters passed are always the same and in the scenario above the ServiceBindContext always have the ITestService as contract and TestService as type even if the method to bind is the method of ITestSubService which lead in a method not found [here](https://github.com/protobuf-net/protobuf-net.Grpc/blob/main/src/protobuf-net.Grpc/Configuration/ServiceBinder.cs#L188) which never calls the `GetCustomAttributes(inherit: true)` of the type implementation.

Attempt to change
`var bindCtx = new ServiceBindContext(serviceContract, serviceType, state, binderConfiguration.Binder);`
to 
`var bindCtx = new ServiceBindContext(typeToBindItsMethods, serviceType, state, binderConfiguration.Binder);`
all tests passed and all services work without issues, but it is correct? We missed something?

Dona278:
Custom attributes used to decorate methods which came from an interface used for "SubService" aren't collected.

With this fix this code works:
```cs
[SubService]
public interface ITestSubService
{
   [Operation]
   ValueTask DoSomethingCommon(CallContext context);
}

[Service]
public interface ITestService : ITestSubService
{
   [Operation]
   ValueTask DoSomethingSpecific(CallContext context);
}

internal class TestService : ITestService
{
   // This method came from "SubService" interface and the "CustomAttribute" now is collected by endpoint metadata
   [CustomAttribute]
   public ValueTask DoSomethingCommon(CallContext context)
   {
      // ...
   }

   [CustomAttribute]
   public ValueTask DoSomethingSpecific(CallContext context)
   {
      // ...
   }
}
```

### DIFF
--- a/src/protobuf-net.Grpc/Configuration/ServerBinder.cs
+++ b/src/protobuf-net.Grpc/Configuration/ServerBinder.cs
@@ -64,7 +64,7 @@ namespace ProtoBuf.Grpc.Configuration
                     var serviceContractSimplifiedExceptions = serviceImplSimplifiedExceptions ||
                                                               typeToBindItsMethods.IsDefined(
                                                                   typeof(SimpleRpcExceptionsAttribute));
-                    var bindCtx = new ServiceBindContext(typeToBindItsMethods, serviceType, state, binderConfiguration.Binder);
+                    var bindCtx = new ServiceBindContext(serviceContract, serviceType, state, binderConfiguration.Binder);
                     foreach (var op in ContractOperation.FindOperations(binderConfiguration, typeToBindItsMethods, this))
                     {
                         if (ServerInvokerLookup.TryGetValue(op.MethodType, op.Context, op.Arg, op.Result, op.Void, out var invoker)
--- a/src/protobuf-net.Grpc/Configuration/ServerBinder.cs
+++ b/src/protobuf-net.Grpc/Configuration/ServerBinder.cs
@@ -64,7 +64,7 @@ namespace ProtoBuf.Grpc.Configuration
                     var serviceContractSimplifiedExceptions = serviceImplSimplifiedExceptions ||
                                                               typeToBindItsMethods.IsDefined(
                                                                   typeof(SimpleRpcExceptionsAttribute));
-                    var bindCtx = new ServiceBindContext(serviceContract, serviceType, state, binderConfiguration.Binder);
+                    var bindCtx = new ServiceBindContext(typeToBindItsMethods, serviceType, state, binderConfiguration.Binder);
                     foreach (var op in ContractOperation.FindOperations(binderConfiguration, typeToBindItsMethods, this))
                     {
                         if (ServerInvokerLookup.TryGetValue(op.MethodType, op.Context, op.Arg, op.Result, op.Void, out var invoker)
--- a/src/protobuf-net.Grpc/Configuration/ServiceBinder.cs
+++ b/src/protobuf-net.Grpc/Configuration/ServiceBinder.cs
@@ -211,7 +211,7 @@ namespace ProtoBuf.Grpc.Configuration
         {
             if (contractType != serviceType & serviceMethod is object)
             {
-                var map = GetMap(contractType, serviceType);
+                var map = GetMap(serviceMethod.DeclaringType ?? contractType, serviceType);
                 var from = map.InterfaceMethods;
                 var to = map.TargetMethods;
                 int end = Math.Min(from.Length, to.Length);
--- a/src/protobuf-net.Grpc/Configuration/ServiceBinder.cs
+++ b/src/protobuf-net.Grpc/Configuration/ServiceBinder.cs
@@ -209,7 +209,7 @@ namespace ProtoBuf.Grpc.Configuration
         /// </summary>
         public MethodInfo? GetMethodImplementation(MethodInfo serviceMethod, Type contractType, Type serviceType)
         {
-            if (contractType != serviceType & serviceMethod is object)
+            if (contractType != serviceType && serviceMethod is object)
             {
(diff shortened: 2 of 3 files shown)


Dona278:
Hi @mgravell! Sorry to take the opportunity, but I've seen the great work you're doing to make the libraries AOT-ready. I've had the fork for two years just for this fix, and I wanted to take advantage of your active presence to ask you about the merge. Thanks!

mgravell:
Thanks for this — the diagnosis is right, and #330 is a genuine bug. `GetInterfaceMap` only lists the methods an interface *itself* declares, so a `[SubService]` operation is never found and the implementation method's attributes are dropped:

```
GetInterfaceMap(ISvc).InterfaceMethods = [Specific]
   contains ISub.Common? NO  -> attributes lost
GetInterfaceMap(ISub).InterfaceMethods = [Common]
   contains ISub.Common? YES -> 1 attr(s) on impl
```

I think the one-line change has a side effect worth catching before it lands, though.

### `ContractType` is doing two jobs

It has exactly one consumer — `GetMetadata(method, ContractType, ServiceType)` — and that method uses it for two unrelated things:

1. `contractType.GetCustomAttributes(inherit: true)` — the contract **type's** attributes;
2. `GetMethodImplementation(method, contractType, serviceType)` → `GetInterfaceMap(contractType)` — to find the implementing method and read **its** attributes.

Only (2) is broken. Changing the parameter fixes (2) but also moves (1), and that matters because **attributes on an interface do not inherit** — `GetCustomAttributes(inherit: true)` does not walk base interfaces:

```
IBase.GetCustomAttributes(inherit:true)    -> base-interface
IDerived.GetCustomAttributes(inherit:true) -> derived-interface
```

So it is a swap rather than a gain. Running the shipped `ServiceBinder.GetMetadata` both ways, for an operation inherited from a `[SubService]` base:

| `[SomeAttribute]` on… | before | after |
| --- | :-: | :-: |
| top-level `[Service]` interface | **gathered** | **lost** |
| `[SubService]` interface | not gathered | gathered |
| implementation method | not gathered | gathered |

Two of those are the fix; the first is a regression. The concrete worry is `[Authorize]` on the top-level service contract: it currently reaches operations inherited from a `[SubService]` base, and afterwards it would not — an endpoint quietly stops requiring authentication, with nothing failing to signal it.

It is easy to miss in review because it only affects sub-service operations (for the contract's own methods `typeToBindItsMethods == serviceContract`, so that path is unchanged), and because it reads like "use the more specific type" when it is really "use a different type".

### A smaller fix that keeps all three

`GetMethodImplementation` already has the method, and `serviceMethod.DeclaringType` **is** the declaring interface — which is exactly what job (2) wants, while job (1) keeps wanting the service contract. So in `ServiceBinder`:

```diff
-                var map = GetMap(contractType, serviceType);
+                var map = GetMap(serviceMethod.DeclaringType ?? contractType, serviceType);
```

and `ServerBinder.cs` needs no change at all. That is correct for the non-sub-service case too, where `DeclaringType == contractType`.

If the `[SubService]` interface's own type-level attributes are also wanted, that is a separate and additive choice — collect from both types — rather than replacing one with the other.

mgravell:
Note crossover with #370

Dona278:
@mgravell thanks! It almost looks like you wrote this library :D I edited the fix and added a test (also I tested the code locally with our services)

mgravell:
This is exactly it — thank you, and thanks for bearing with the back-and-forth to get here.

The thing I was worried about was losing the top-level contract type's attributes while gaining the
implementation's, and the test settles that rather than leaving me to reason about it: every one of the
three expectations *starts* with `ContractType`, so nothing is traded away. Covering the explicit
interface implementation (`void IIssue330Sub.ExplicitSub()`) is the case I'd have most expected a map
lookup to fall down on, so I'm glad it's pinned.

`&` → `&&` is a nice catch while you were in there.

I ran it locally across the TFMs before merging — 197/197 on net8.0 and net10.0, 195/195 on net472.

Worth recording for anyone who finds this later: the sub-service interface's own *type-level* attribute
is still not collected, which is consistent with `GetCustomAttributes(inherit: true)` not walking base
interfaces, so it's existing semantics rather than something this introduces. Widening that would be a
separate behaviour change, and I prefer that the test pins today's behaviour explicitly over leaving it
ambiguous.

This also composes well with the build-time gRPC proxies work in protobuf-net, which passes the
*declaring* interface's `MethodInfo` — so generated bindings pick this fix up with no change at all.

